PR #694, opened 2022-10-03
Fix empty chunk sections

KCodeYT:


Superice666 (changes requested):
In many cases, chunk sections also need to preserve biome data. For example, the biomes in nether.

KCodeYT:
> In many cases, chunk sections also need to preserve biome data. For example, the biomes in nether.

But then they are not empty they contain biome data

Superice666:
> But then they are not empty they contain biome data

I think that empty just means that they contain no blocks. Containing biome data is necessary for each ChunkSection.

KCodeYT:
> > But then they are not empty they contain biome data
> 
> I think that empty just means that they contain no blocks. Containing biome data is necessary for each ChunkSection. Just imagine that you are build a high building in the_end, and when you get to its top, finding that the sun is shining because the empty chunksection contains no biome data.

Hmm I think the copy last flag should be used for empty chunk sections

PR #695, opened 2022-10-03
Fix biomes

KCodeYT:


Superice666 (approved):

src/main/java/cn/nukkit/level/format/generic/BaseChunk.java
```diff
@@ -221,7 +221,7 @@ public boolean setBlockStateAtLayer(int x, int y, int z, int layer, BlockState s
         }
     }
 
-    private ChunkSection getOrCreateMutableSection(int sectionY) {
+    protected ChunkSection getOrCreateMutableSection(int sectionY) {
```
Superice666, line 224: Copy the biome data in the old chunk section when creating a new chunk section.

Superice666 (changes requested):

PR #958, opened 2023-01-24
Better Discord

IndianBartonka:


Superice666:
It will be better if we sync this PR to both 简体中文 and 繁體中文 version.

PR #1138, opened 2023-04-12
Refactoring of player name checking & issue #486 fix

KoshakMineDEV:
Fixed issue #486

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -426,7 +426,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
 
     @PowerNukkitXOnly
     @Since("1.19.70-r3")
-    public Pattern playerNameRegex = Pattern.compile("^(?! )([a-zA-Z0-9_ ]{2,16}[a-zA-Z0-9_])(?<! )$");
+    private static final Pattern playerNameRegex = Pattern.compile("^(?! )([a-zA-Z0-9_ ]{2,16}[a-zA-Z0-9_])(?<! )$");
 
 
     /**
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -127,6 +127,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static cn.nukkit.utils.Utils.dynamic;
@@ -422,6 +423,10 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
     protected Entity lastBeAttackEntity = null;
 
     private boolean foodEnabled = true;
+
+    @PowerNukkitXOnly
+    @Since("1.19.70-r3")
+    public Pattern playerNameRegex = Pattern.compile("^(?![_ ])([a-zA-Z0-9_]{3,16}(?: [a-zA-Z0-9_]{3,17})*)(?<![_ ])$");
 
     /**
      * 单元测试用的构造函数
@@ -3543,7 +3548,7 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
 
                     boolean valid = true;
 
-                    Matcher usernameMatcher = this.server.playerNameRegex.matcher(loginPacket.username);
+                    Matcher usernameMatcher = this.playerNameRegex.matcher(loginPacket.username);
                     if (!usernameMatcher.matches()) {
                        valid = false;
                     }
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -426,7 +426,8 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
 
     @PowerNukkitXOnly
     @Since("1.19.70-r3")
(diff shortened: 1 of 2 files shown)


src/main/java/cn/nukkit/Player.java
```diff
@@ -3546,6 +3546,10 @@ public void handleDataPacket(DataPacket packet) {
                         valid = false;
                     }
 
+                    if(loginPacket.username.startsWith(" ") || loginPacket.username.startsWith(".")) {
```
Superice666, line 3549: considering using `.isBlank` instead of `.startsWith(" ")`.
Also, some languages write right to left and it's common to have a space or dot at the end of one's name in some cultures.

Superice666 (changes requested):

src/main/java/cn/nukkit/Player.java
KoshakMineDEV: I think it's better to completely redo the nickname check with regex. 
In minecraft without xbox you can use most of unicode symbols. But in xbox live and nickname required to follow some rules:
Your nickname not started with space, dot
nickname not ended with space or dot
And also you can use only Latin alphabet

Why i recommend to use xbox live like nickname check system? Because with this system plugins cannot be broken in case of weird nickname.

Superice666 (changes requested):

src/main/java/cn/nukkit/Server.java
```diff
@@ -347,6 +347,10 @@ public Level remove(Object key) {
     @Since("1.19.50-r1")
     private FreezableArrayManager freezableArrayManager;
 
+    @PowerNukkitXOnly
+    @Since("1.19.70-r3")
+    public Pattern playerNameRegex = Pattern.compile("^(?![_ ])([a-zA-Z0-9_]{3,16}(?: [a-zA-Z0-9_]{3,17})*)(?<![_ ])$");
```
Superice666, line 352: Now names with a space in the middle like `Joe Biden` are not allowed, however it should be.
Maybe we lost a space in the charset `[a-zA-Z0-9_]`?

Superice666 (changes requested):

src/main/java/cn/nukkit/Server.java
KoshakMineDEV: Name with space in it perfectly works, my nickname is "koshak mine" and i can join

Superice666: Sorry for giving an incorrect example earlier. It seems that a two character surname is not allowed. But this is at least common in Chinese names translated into English. Can you make some modifications to this? esteem it a favor

KoshakMineDEV: If I understand you correctly, it should work now

Superice666: thanks, LGTM.
I'm requesting for at least one more review, if everything is ok, this PR will be merged.

Superice666 (approved):

src/main/java/cn/nukkit/Player.java
```diff
@@ -422,6 +424,11 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
 
     private boolean foodEnabled = true;
 
+    @PowerNukkitXOnly
+    @Since("1.19.70-r3")
+    public Pattern playerNameRegex = Pattern.compile("^(?! )([a-zA-Z0-9_ ]{2,16}[a-zA-Z0-9_])(?<! )$");
```
CoolLoong, line 429: change visibility to private

CoolLoong (changes requested):

src/main/java/cn/nukkit/Player.java
CoolLoong: and static final

src/main/java/cn/nukkit/Player.java
```diff
@@ -3541,23 +3548,10 @@ public void handleDataPacket(DataPacket packet) {
                     this.rawUUID = Binary.writeUUID(this.uuid);
 
                     boolean valid = true;
-                    int len = loginPacket.username.length();
-                    if (len > 16 || len < 3) {
```
CoolLoong, line 3545: Do not remove this length detection

src/main/java/cn/nukkit/Player.java
```diff
@@ -3541,23 +3548,10 @@ public void handleDataPacket(DataPacket packet) {
...
-                                (c >= 'A' && c <= 'Z') ||
-                                (c >= '0' && c <= '9') ||
-                                c == '_' || c == ' '
-                        ) {
-                            continue;
-                        }
 
-                        valid = false;
-                        break;
+                    Matcher usernameMatcher = this.playerNameRegex.matcher(loginPacket.username);
```
CoolLoong, line 3552: remove `this.`

src/main/java/cn/nukkit/Player.java
```diff
@@ -422,6 +424,11 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
 
     private boolean foodEnabled = true;
 
+    @PowerNukkitXOnly
+    @Since("1.19.70-r3")
+    private static final Pattern playerNameRegex = Pattern.compile("^(?! )([a-zA-Z0-9_ ]{2,16}[a-zA-Z0-9_])(?<! )$");
```
CoolLoong, line 429: add annotation what username will this reject

CoolLoong (changes requested):

src/main/java/cn/nukkit/Player.java
KoshakMineDEV: there is lenght detection in regex

CoolLoong (approved):

smartcmd (approved):
It would be better if you can add a small unit test

PR #1271, opened 2023-06-28
Runtime-only entities

alvin0319:
After this PR, entities marked as not saving with chunk won't be saved into the disk.

### DIFF
--- a/src/main/java/cn/nukkit/entity/Entity.java
+++ b/src/main/java/cn/nukkit/entity/Entity.java
@@ -578,8 +578,8 @@ public abstract class Entity extends Location implements Metadatable {
     private volatile boolean initialized;
 
     @PowerNukkitXOnly
-    @Since("1.20.0-r2")
-    protected boolean saveWithChunk = true;
+    @Since("1.20.0-r3")
+    protected volatile boolean saveWithChunk = true;
 
     public Entity(FullChunk chunk, CompoundTag nbt) {
         if (this instanceof Player) {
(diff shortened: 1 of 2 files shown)


src/main/java/cn/nukkit/entity/Entity.java
```diff
@@ -577,6 +577,10 @@ public abstract class Entity extends Location implements Metadatable {
     private int maxHealth = 20;
     private volatile boolean initialized;
 
+    @PowerNukkitXOnly
+    @Since("1.20.0-r2")
+    protected boolean saveWithChunk = true;
```
SuperIceCN, line 582: it should be volatile since server may crash and save the chunk while the entity object is constructing.

src/main/java/cn/nukkit/entity/Entity.java
```diff
@@ -577,6 +577,10 @@ public abstract class Entity extends Location implements Metadatable {
     private int maxHealth = 20;
     private volatile boolean initialized;
 
+    @PowerNukkitXOnly
+    @Since("1.20.0-r2")
```
SuperIceCN, line 581: since 1.20.0-r3, r2 is already released.

src/main/java/cn/nukkit/level/format/anvil/Chunk.java
```diff
@@ -516,6 +516,7 @@ public byte[] toFastBinary() {
     }
 
     @Override
+    @PowerNukkitXDifference(since= "1.20.0-r2", info = "Do not save entities which marked as not savable with chunk.")
```
SuperIceCN, line 519: since 1.20.0-r3

src/main/java/cn/nukkit/entity/Entity.java
```diff
@@ -1977,6 +1981,14 @@ public boolean canCollideWith(Entity entity) {
         return !this.justCreated && this != entity && !this.noClip;
     }
 
+    @PowerNukkitXOnly
+    @Since("1.20.0-r2")
```
SuperIceCN, line 1985: since 1.20.0-r3

src/main/java/cn/nukkit/entity/Entity.java
```diff
@@ -1977,6 +1981,14 @@ public boolean canCollideWith(Entity entity) {
...
+    public boolean canBeSavedWithChunk() { return saveWithChunk; }
+
+    @PowerNukkitXOnly
+    @Since("1.20.0-r2")
```
SuperIceCN, line 1989: since 1.20.0-r3

SuperIceCN (changes requested):

SuperIceCN (approved):

PR #1306, opened 2023-07-08
fix: vanilla command lang of `fra`

CoolLoong:


AzaleeX (approved):
after checking it's good for me I don't see any grammar problem

PR #1310, opened 2023-07-11
Offload packet encoding and decoding from main thread

SuperIceCN:
In situations where there are many entities, players, or particles, a large number of data packets are all piled up on the main thread for encoding and decoding, which will seriously affect performance.

This PR adds PacketProcessingThread. Most data packets will be encoded and decoded using this Thread. According to the test, the load of the main thread can be reduced by 30% under the condition of about 100 entities+players.

Because the MemoryBarriers is used to synchronize packet data between multiple threads, the performance on single core and dual core CPUs decreases slightly.

More measured results are needed to demonstrate the absence of potential concurrent errors.

-------

Bugs:
- [x] EntityMetaData does not support concurrent access.
- [x] Inbound datapacket sometimes are parsed in wrong orders.
- [x] Outbound datapacket sometimes are sent in wrong orders.
- [x] Concurrent encrypting is not supported.

### DIFF
--- a/src/main/java/cn/nukkit/network/session/RakNetPlayerSession.java
+++ b/src/main/java/cn/nukkit/network/session/RakNetPlayerSession.java
@@ -114,7 +114,7 @@ public class RakNetPlayerSession implements NetworkPlayerSession, RakNetSessionL
             buffer.readBytes(packetBuffer);
 
             try {
-                this.server.getNetwork().processBatch(packetBuffer, this.inbound, this.getCompression());
+                this.server.getNetwork().processBatch(packetBuffer, this.inbound, this.getCompression(), getPacketProcessingExecutor());
             } catch (ProtocolException e) {
                 this.disconnect("Sent malformed packet");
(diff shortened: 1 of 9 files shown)


AzaleeX:
I believe the maven build doesn't like what you did 🤣

SuperIceCN:
> I believe the maven build doesn't like what you did 🤣

Now it likes this PR. 😎

src/main/java/cn/nukkit/network/session/RakNetPlayerSession.java
```diff
@@ -27,6 +28,7 @@
 import it.unimi.dsi.fastutil.objects.ObjectArrayList;
 import lombok.extern.log4j.Log4j2;
 import org.apache.logging.log4j.message.FormattedMessage;
+import org.jetbrains.annotations.NotNull;
```
AzaleeX, line 31: I don't think this is useful

CoolLoong (approved):

PR #1325, opened 2023-07-21
Fixed wrong damagecause

Buddelbubi:
Thons damage had the damage reason ENTITY_ATTACK instead of its decicated THORNS cause. This made it nearly impossible to detect thorns damage.


smartcmd (approved):

smartcmd:
Thanks for your contribution!

AzaleeX:
well seen I had it in the back of my mind but I didn't have time to fix it but in any case thank you for the contribution

Issue #1328, opened 2023-07-22
Dead Bush Update


AzaleeX:

AzaleeX:
@SuperIceCN the sweater has been checked so I can merge it or I have to wait for you to do it

PR #1329, opened 2023-07-22
Update of the Dead Bush so that it can be placed on grass or moss

AzaleeX:
Updated Dead bush so it can be placed on moss block and grass which was not possible before

https://github.com/PowerNukkitX/PowerNukkitX/issues/1328


AzaleeX:
https://github.com/PowerNukkitX/PowerNukkitX/issues/1225

smartcmd (approved):

PR #1331, opened 2023-07-23
fixing /fill replace logic

Sherko231:
before : when you do /fill xyz xyz concrete 5 replace concrete  ..... it was replacing only concrete that it's data = 0 (in this case ... only white concrete)
now : when you do the same command .... it will replace all type of concretes with the block you want (same to vanilla minecraft)


smartcmd:
Thanks for your contributions

lt-name (approved):

PR #1333, opened 2023-07-23
added bamboo boat

AzaleeX:
I just have something that I don't understand is that the item does not go in the /give nor in the creative menu


AzaleeX:
⚠I just have something that I don't understand is that the item does not go in the /give nor in the creative menu

smartcmd:
Thanks for your contributions! :heart:

lt-name (approved):

PR #1336, opened 2023-07-23
/fill hollow-outline fix

Sherko231:
before : when doing `/fill xyz xyz glass 0 hollow  or /fill xyz xyz glass 0 outline `... it would work only if the coordinates were only numbers (wasn't working for ~ or ^)

after : when you do :` /fill ~-4 ~-4 ~-4 ~4 ~4 ~4 glass 0 hollow `.... it will work ! same for outline


smartcmd:
Thanks again! :heart:

Sherko231:
np ❤ happy to help

PR #1339, opened 2023-07-30
Separate plugin authors with ", " in the version command

xxFLORII:
This change separates authors so that instead of "Authors: xxFLORIIxxAROX" it looks like "Authors: xxFLORII, xxAROX".


AzaleeX (approved):

AzaleeX:
Thank you for your contribution 🎉

PR #1351, opened 2023-08-07
Block Entity Bug fixed!

Buddelbubi:
I guess I finally made it... It was so.. It took at least 6 hours of testing and fixing.. 
But at some point, I managed to get rid of the **Could not create a block entity of type {} with x args** message by finding the cause and fixing it.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -6149,13 +6149,13 @@ public class Player extends EntityHuman implements CommandSender, InventoryHolde
         }
     }
     @PowerNukkitXOnly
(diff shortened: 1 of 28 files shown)


Buddelbubi:
I swear I will create a new fork after this...

AzaleeX (commented):
for now it seems ok to me if you add something / an extra commit if you can mention me 😎

CoolLoong (approved):

smartcmd:
It looks like there are some backward-breaking changes. Should have more tests

Buddelbubi:
What should I look for? I installed this on syodo to check if there are any problems in the long terms.

PR #1354, opened 2023-08-09
Initial Support for Smithing Template

glorydark:
## Intro
只是初步的支持，id从802开始注册（走弯路走太久了，累了awa，就帮到这了）
In this PR, only items are registried and are ensured to display normally like the snapshot below.
Thanks to Geyser's sourcecode, the trimmed armor are able to display in the server.

## Snapshot
QQ截图20230810003613

## Reference
https://github.com/davchoo/Geyser/blob/09d9b54afd73afd2f2df39d3a729475edacf980e/core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaUpdateRecipesTranslator.java#L48

## Others
如果你想要使用此功能，可以参照如下的代码（ItemArmorPatternType.java和ItemMaterialType.java能方便你选择对应的类型）
If you want to implement template to an armor, here is a simple example I used to test this PR. (ItemArmorPatternType.java & ItemMaterialType.java might help you.)
```java
Player player = sender.asPlayer();
Item item = player.getInventory().getItemInHand().clone();
CompoundTag tag = new CompoundTag();
if(item.hasCompoundTag()){
    tag = item.getNamedTag();
}
tag = tag.putCompound(new CompoundTag("Trim")
               .putString("Pattern", "coast")
               .putString("Material", "gold"));
item.setCompoundTag(tag);
player.getInventory().setItemInHand(item);
```


AzaleeX:
amazing to implement this in the future I will make an interface with all armor models so that the api is more correct, thanks for the contribution ❤

glorydark:
也许我们需要做**更多的改动**，这样玩家就可以在锻造台界面看到使用纹饰装饰的效果，然后选择是否装饰他们的装甲。我已经尝试了几次调整，但不幸的是，仍然有一些错误（**点击拿起合成物就会关闭GUI**）。尽力了awa。
Maybe we need to do **more changes**, so that in the GUI of Smithing Table, player can see the trimmed result and then choose to trim their armors. I have made some tweaks on my local branch, but unfortunately, there's still something wrong(**The GUI always automatically close upon picking up the result item**). I've tried my best. 

但至少有几点是可以确定的：
But fortunately, there are still some facts to be sure.

## Change on SmithingInventory.class
现在锻造库存有**三个**槽位放置材料或盔甲。但是在smithinginventory .类中，它只支持两个槽位，这可以追溯到 [20w17a / 1.19.4](https://minecraft.fandom.com/zh/wiki/20w17a)
Now smithing inventory has **three** slots for putting ingredients or armors. But in the SmithingInventory.class, it only supports two slots, which can be traced back to [20w17a / 1.19.4](https://minecraft.fandom.com/zh/wiki/20w17a)
```java
    private static final int EQUIPMENT = 0;
    private static final int INGREDIENT = 1;
    private static final int TEMPLATE = 2; // We need this.
```
| OLD | NEW |
| --- | --- |
| Old Smithing UI | New Smithing UI |

## Change on CraftingDataPacket.class
下面的源代码可以向玩家显示结果，但玩家无法取出输出物品。
The sourcecode below can help display the result item to the player, but player still **cannot** use it to trim their armours.
```java
@Override
    public void encode() {
        this.reset();
        this.putUnsignedVarInt(entries.size() + 1);
        ...
        for (Recipe recipe : entries) {
            this.putVarInt(recipe.getType().networkType);
            switch (recipe.getType()) {
                ...
                case SMITHING_TRIM -> {
                    // no-op
                }
            }
        }

        // Identical smithing_trim recipe sent by BDS that uses tag-descriptors, as the client seems to ignore the
        // approach of using many default-descriptors (which we do for smithing_transform)
        this.putVarInt(RecipeType.SMITHING_TRIM.networkType);
        this.putString("minecraft:smithing_armor_trim");
        this.putRecipeIngredient(new ItemTagDescriptor("minecraft:trim_templates", 1));
        this.putRecipeIngredient(new ItemTagDescriptor("minecraft:trimmable_armors", 1));
        this.putRecipeIngredient(new ItemTagDescriptor("minecraft:trim_materials", 1));
        this.putString(CRAFTING_TAG_SMITHING_TABLE);
        this.putUnsignedVarInt(recipeNetworkId);
        ...
    }
```

## Smithing(Upgrade)Recipe
在没有任何改变的情况下，锻造升级配方在游戏中运行良好。玩家可以升级钻石装备为合金装备。
Without any change, **original smithing upgrade recipes work well in games**. Player can upgrade upgrade diamond equipment with netherite.

glorydark:
name已经开始对锻造模板进行进一步支持，建议移步关注：https://github.com/MemoriesOfTime/Nukkit-MOT/pull/2

I've not surpassed the problem in the process of further support. Fortunately, Nukkit-MOT has created a PR draft that is connected with this PR. You can tune in that PR for following support and changes.

PR #1363, opened 2023-08-22
Added support to shut down a server with ctrl+c

AzaleeX:


lt-name (approved):

src/main/java/cn/nukkit/utils/ServerKiller.java
```diff
@@ -19,13 +21,17 @@ public ServerKiller(long time, TimeUnit unit) {
     }
 
     @Override
+    @PowerNukkitOnly
```
lt-name, line 24: This note doesn't seem to be needed

Issue #1370, opened 2023-09-02
Missing crafting recipes


tunikakeks:
# 🐞 I found a bug

### 📸 Screenshots / Videos
Example recipes:
![image](https://github.com/PowerNukkitX/PowerNukkitX/assets/54219265/01e47a82-48dd-497e-a238-ac3c1e69e753)
![image](https://github.com/PowerNukkitX/PowerNukkitX/assets/54219265/0be2f985-879c-4a19-ba49-495853fb3096)
![image](https://github.com/PowerNukkitX/PowerNukkitX/assets/54219265/a58e8123-3354-4d53-b412-f41dd14eab88)

### ▶ Steps to Reproduce
Try to craft wooden logs into planks / Try to craft quartz pillars

### ✔ Expected Behavior
Quartz pillars would get crafted insteadof another quartzblocks &
Planks should be the same type as the log (Spruce -> Spruce)

### ❌ Actual Behavior
The quartz pillar recipe gives you 2 quartzblocks and planks aren't the same type, most of them are resulting in oak (Spruce -> Oak)

* PowerNukkit version: 1.20.10-r1

Verox001:
Thanks for the report. Will fix it in the next pull request.

AzaleeX:
I'm thinking of working on this `craft problem` **tomorrow**.

Verox001:
@tunikakeks Unfortunately I cannot confirm that crafting quartz pillars does result in two quartz blocks. Please provide more information if this problem persists for you or try updating PNX.
The bug with wrong wood types is fixed in the PR above.

tunikakeks:
@Verox001 I just tried your fix but it don't work on my end. I used your changes to build pnx and started my server with it
![image](https://github.com/PowerNukkitX/PowerNukkitX/assets/54219265/a40a4d06-851d-41f3-a57c-ad902aa3702d)

Verox001:
You probably compiled a wrong version or something like that. Let someone accept my PR first. And then make sure there is nothing wrong with your server (maybe you're starting it from a wrong entry file).

tunikakeks:
I just checked it, it's weird. On a new server it works fine, but when I try it on my old server with the updated PNX.jar, it doesnt work (its the same jar file and the same start process)

Verox001:
Yes, this seems more like a problem only occuring to you. @smartcmd May you be so kind accepting the PR and then marking this issue as solved, since everything else are probably deprecated errors?

[...]

PR #1371, opened 2023-09-02
Fixing the recipe conventions for block_states

Verox001:
(OLD_LOG_TYPE was deprecated)

Changes:
- Made BlockWood use the same WoodType Block state as the other blocks


Verox001:
Not finished yet, but I will fix #1370 with this PR.

Verox001:
Done for now.
Please review and merge to fix the mentioned issue.

smartcmd:
Will this cause backward compatibility issues?

Verox001:
No, since the tags for the items are compiled as well and therefore will cause no issues, when updating.

smartcmd (approved):

AzaleeX (approved):

PR #1377, opened 2023-09-14
WaterdogPE fix

Buddelbubi:
Undid StartGamePacket Changes made in ([#1357](https://github.com/PowerNukkitX/PowerNukkitX/pull/1357))
The protocol library used in WaterdogPE cannot deserialize the custom data. 
Custom Blocks and items are working anyways without sending thie data in the StartGamePacket.

grafik

PR #1378, opened 2023-09-17
Waterdog fix fix for non wdpe servers

Buddelbubi:
Somehow those changes were not commited in the first time. I added them back and tested again for WDPE and non WDPE Servers.

### DIFF
--- a/src/main/java/cn/nukkit/network/protocol/StartGamePacket.java
+++ b/src/main/java/cn/nukkit/network/protocol/StartGamePacket.java
@@ -226,7 +226,6 @@ public class StartGamePacket extends DataPacket {
         this.putString(""); // EduSharedUriResource linkUri
         if (Server.getInstance().isEnableExperimentMode() && !Server.getInstance().isWaterdogCapable()) { // force Experimental Gameplay
             this.putBoolean(true); // force Experimental Gameplay
-            this.putBoolean(true);
         } else {
             this.putBoolean(false);
         }
--- a/src/main/java/cn/nukkit/network/protocol/StartGamePacket.java
+++ b/src/main/java/cn/nukkit/network/protocol/StartGamePacket.java
@@ -224,9 +224,9 @@ public class StartGamePacket extends DataPacket {
         this.putBoolean(false); // Nether type
         this.putString(""); // EduSharedUriResource buttonName
         this.putString(""); // EduSharedUriResource linkUri
-        if (Server.getInstance().isEnableExperimentMode()) { // force Experimental Gameplay
-            this.putBoolean(Server.getInstance().isEnableExperimentMode()); // force Experimental Gameplay
-            this.putBoolean(!Server.getInstance().getConfig("settings.waterdogpe", false)); // Why WaterDogPE require an extra optional boolean if this is set to true? I don't know.
+        if (Server.getInstance().isEnableExperimentMode() && !Server.getInstance().isWaterdogCapable()) { // force Experimental Gameplay
+            this.putBoolean(true); // force Experimental Gameplay
+            this.putBoolean(true);
         } else {
             this.putBoolean(false);
         }
--- a/src/main/java/cn/nukkit/network/protocol/StartGamePacket.java
+++ b/src/main/java/cn/nukkit/network/protocol/StartGamePacket.java
@@ -183,7 +183,7 @@ public class StartGamePacket extends DataPacket {
         this.putBoolean(this.commandsEnabled);
         this.putBoolean(this.isTexturePacksRequired);
         this.putGameRules(this.gameRules);
-        if (Server.getInstance().isEnableExperimentMode()) {
+        if (Server.getInstance().isEnableExperimentMode() && !Server.getInstance().isWaterdogCapable()) {
             this.putLInt(4); // Experiment count
             {
                 this.putString("data_driven_items");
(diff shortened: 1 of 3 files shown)


Mcayear:
nukkit.yml works fine in `waterdogpe: true` configuration, but still cannot enter the server in `waterdogpe: false`.

Buddelbubi:
Wait, I'll try again

Buddelbubi:
Are you sure that you compiled the right code? (waterdog-fix branch)
I tested again with waterdogpe set to false as well with enabled and disabled experiment mode. Always was able to join.

Mcayear:
use the action Artifacts: https://github.com/PowerNukkitX/PowerNukkitX/actions/runs/6214928045#Artifacts

Previous pr#1377 could not join the server regardless of whether `waterdogpe` was enabled in nukkit.yml or not.

No plug-ins installed, only pure powernukkitx server is used.

Buddelbubi:
I cannot start it. For some reason it tells me to use the CLI. The self compiled shaded jar starts as normal... Any Ideas? I enabled allow-shaded so it should start..

Buddelbubi:
> nukkit.yml works fine in `waterdogpe: true` configuration, but still cannot join the server in `waterdogpe: false`.
> 
> 63b6cc3383bc61533e518caba638ebf8

Okay, I used the build from artifacts and cannot confirm your problem. I am still able to join with waterdog disabled.

Mcayear:
> I cannot start it. For some reason it tells me to use the CLI. The self compiled shaded jar starts as normal... Any Ideas? I enabled allow-shaded so it should start..
> 
> Nvm. https://github.com/PowerNukkitX/PowerNukkitX/blob/9190b0bec837fd7bc8e3397bc754f42980978fe7/src/main/java/cn/nukkit/utils/StartArgUtils.java#L22C9-L24C10 It checks for the file name to contain "shaded"

Download [PowerNukkitX-Core](https://github.com/PowerNukkitX/PowerNukkitX/suites/16259803365/artifacts/928587586) and [PowerNukkitX-Libs](https://github.com/PowerNukkitX/PowerNukkitX/suites/16259803365/artifacts/928587589), and use the [Run from command](https://github.com/PowerNukkitX/PowerNukkitX#2run-from-command) method to run.

Mcayear:
> Okay, I used the build from artifacts and cannot confirm your problem. I am still able to join with waterdog disabled.

I've set up a test server, want to give it a try?
IP: tale.0mc.me Port: 11006

Buddelbubi:
Are you sure that you are using the correct file? Its contradicting... It works for me...

Buddelbubi:
https://github.com/PowerNukkitX/PowerNukkitX/commit/761f7e3506dfec47f87839bbb54c4a4708b22eec

Mcayear:
> Are you sure that you are using the correct file? Its contradicting... It works for me... grafik
> 
> All I did was basically revert some changed lt-name made before. It makes no sence that the same code just stops working..

wow, maybe i need update to the latest version of client.

i try again.

Mcayear:
1.20.10~1.20.15 is the same result.

Only change server configuration file (nukkit.yml & server.properties):

| Experiment Mode | Water Dog | available |
|--------|--------|--------|
| on | on | ✓ |
| on | off | X |
| off | on | ✓ | 
| off | off | ✓ |

Mcayear:
We'll use protocol 582 as an example.

I noticed that you added an extra line at:
https://github.com/PowerNukkitX/PowerNukkitX/blob/f41ac6772ece9a2927597fa86fae6ba070d4fc7e/src/main/java/cn/nukkit/network/protocol/v582/StartGamePacket_v582.java#L222

You added this line:
this.putBoolean(true);

But in the surrounding if logic, you have `!Server.getInstance().isWaterdogCapable()`, which causes an extra boolean to be sent when Waterdog is false. This is **not equivalent** to the previous lt_name code:

https://github.com/Buddelbubi/PowerNukkitX/blob/761f7e3506dfec47f87839bbb54c4a4708b22eec/src/main/java/cn/nukkit/network/protocol/v582/StartGamePacket_v582.java#L222

Buddelbubi:
> 1.20.10~1.20.15 is the same result.
> 
> Only change server configuration file (nukkit.yml & server.properties):
> Experiment Mode 	Water Dog 	available
> on 	on 	✓
> on 	off 	X
> off 	on 	✓
> off 	off 	✓

I made the same test. it was working all the time. I did those tests multiple times

Buddelbubi:
Try now when its completed. I removed the extra line. I was sure that it belogs there because I just replaced the content to true since it always results in true because of the if clause.

Mcayear:
> Try now when its completed. I removed the extra line. I was sure that it belogs there because I just replaced the content to true since it always results in true because of the if clause.

tested. It's working fine now.

> Still I cannot explain why it did not work for you in the first place. It was working for me..

lt_name also has this problem.

https://github.com/PowerNukkitX/PowerNukkitX/pull/1377#issuecomment-1721269910

Mcayear:
But your judgment statement will cause the experimental mode to be turned off when waterdog is turned on. because `!Server.getInstance().isWaterdogCapable()` eques `!true`.

I don't know if this is your purpose. x )

Mcayear (approved):

CoolLoong:
Write true twice because the value is of type optional, and the PNX implementation is error actually.

CoolLoong:
After test, I think the implementation of PNX is correct

PR #1381, opened 2023-09-21
feat: 1.20.30

CoolLoong:


AzaleeX (approved):
Sorry I'm late looking at your pull request, but thanks for contributing to PowerNukkitX. ❤

AzaleeX:
Then it still has a problem when it tries to build maven it makes an error.
https://github.com/PowerNukkitX/PowerNukkitX/actions/runs/6257058005/job/16988838022#step:5:16749

PR #1392, opened 2023-09-30
Fix EntityIntelligent and EmptyBehaviorGroup from causing NPEs

FlamingKnight:
Hello! Attacking a hostile mob causes an NPE to be thrown. Spawn something like a guardian and try attacking it. cn.nukkit.entity.EntityMob on line 126 tries to make a mob "hate" the player upon being attacked, however, all mobs implement the EmptyBehaviorGroup class which returns a null IMemoryStorage making the server not have a good time. I don't know how the AI is implemented in the background, though here's a PR that I made that seems to fix the issue, attacking a hostile mob no longer throws an NPE. An interesting note I saw is that it looked like by default mobs should have no AI, however they do, it's not directly related to this though something I realized was weird. If there are any problems with it, let me know!


smartcmd (approved):

smartcmd:
Thanks for your contributions!

FlamingKnight:
Now ready for it to be approved / merged, just needed to change that last object (MemoryStorage -> IMemoryStorage)

PR #1397, opened 2023-10-13
Fixing Directions for Logs

Verox001:
This seems to be general problem with Blocks using BlockAxis with META above 4. For example this problem affects all axis_directions, with a META of 5 - 11.
I just debugged and fixed it that way. It would be nice if someone could show me where the Damage/META of the block is set when changing the Axis.
image


src/main/resources/block_mappings.json
```diff
@@ -132,14 +132,14 @@
...
-    "10": "minecraft:birch_log;pillar_axis=z",
-    "11": "minecraft:jungle_log;pillar_axis=z"
+    "8": "minecraft:oak_log;pillar_axis=x",
+    "9": "minecraft:spruce_log;pillar_axis=x",
+    "10": "minecraft:birch_log;pillar_axis=x",
+    "11": "minecraft:jungle_log;pillar_axis=x",
+    "16": "minecraft:oak_log;pillar_axis=z",
+    "17": "minecraft:spruce_log;pillar_axis=z",
+    "18": "minecraft:birch_log;pillar_axis=z",
+    "19": "minecraft:jungle_log;pillar_axis=z"
```
AzaleeX, line 142: why does this not do the sequence of numbers correctly it's really weird minecraft you can't find

CoolLoong (approved):

PR #1399, opened 2023-10-20
Updated Release Version to 1.20.30-r2

Verox001:
This release features the fix for all block attributes and should resolve the confusion about the latest release having so many bugs.


smartcmd (approved):

CoolLoong:
you need add +u in last commit info

PR #1413, opened 2023-11-04
feat: ability to change Minecraft version in query

KoshakMineDEV:


BlusteryS (commented):
```java
KVdata.put("version", "v" + this.version);
```

It would be good if there was no "v" prefix when setting the custom version.

src/main/java/cn/nukkit/event/server/QueryRegenerateEvent.java
```diff
@@ -107,6 +108,14 @@ public void setPlayerList(Player[] players) {
         this.players = players;
     }
 
+    public String getVersion() {
```
SuperIceCN, line 111: Please add `@PowerNukkitXOnly` and `@Since` here

src/main/java/cn/nukkit/network/RakNetInterface.java
```diff
@@ -164,7 +165,7 @@ public void setName(String name) {
                 .add("MCPE")
                 .add(motd)
                 .add(Integer.toString(ProtocolInfo.CURRENT_PROTOCOL))
-                .add(ProtocolInfo.MINECRAFT_VERSION_NETWORK)
+                .add(dynamic(info.getVersion()))
```
SuperIceCN, line 168: `@PowerNukkitXDifference` here pls

SuperIceCN (changes requested):
LGTM

BlusteryS:
the default value should contain "v" prefix

KoshakMineDEV:
> the default value should contain "v" prefix

We don't need to add it manually, because Minecraft doing it automatically. 
image
This is how it looks in the game with current implementation

SuperIceCN (approved):

PR #1428, opened 2023-11-27
fix: arrow does not disappear

KoshakMineDEV:
Thanks to Nukkit-MOT development team for this fix


smartcmd:
Copy fromnk- mot?

MEFRREEX:
> Copy from nk- mot?

yes, it's from nukkit-mot

smartcmd:
It would be better if lt-name open this pul request

CoolLoong (approved):

PR #1429, opened 2023-11-28
fix: ore & stone types generation

KoshakMineDEV:


src/main/java/cn/nukkit/level/generator/object/ore/OreType.java
```diff
@@ -88,7 +88,7 @@ public boolean spawn(ChunkManager level, NukkitRandom rand, int replaceId, int x
                                 double zVal = ((double) zSeg + 0.5D - scaleZ) / (randVec1 / 2.0D);
 
                                 if (xVal * xVal + yVal * yVal + zVal * zVal < 1.0D) {
-                                    if (level.getBlockIdAt(xSeg, ySeg, zSeg) != 0) {
+                                    if (level.getBlockIdAt(xSeg, ySeg, zSeg) == STONE) {
```
CoolLoong, line 91: why?Other types of stones should also be replaced,should not only replace stone

KoshakMineDEV: This types of stone is also generated by TypeOre. Currently, after @Buddelbubi world generator changes ores can spawn on the terrrain and in the water.

KoshakMineDEV: only one thing i need to check is ore generation in deeplstate

KoshakMineDEV:
Now works 100% right

CoolLoong (approved):

src/main/java/cn/nukkit/level/generator/object/ore/OreType.java
Buddelbubi: Oh sorry. Didn't thought about that

PR #1437, opened 2023-12-08
1.20.50 Modifications from KoshakMineDEV

lt-name:


BlusteryS:
Screenshot_2023-12-09-09-26-38-737_com android chrome
Why?

KoshakMineDEV:
1.20.40-r2 was NEVER released. That's why I changed it to 1.20.50-r1

PR #1449, opened 2023-12-29
refactor: adapt some BlockStem and BlockCropsStem

Mcayear:


Mcayear:
There may be a few omissions.

CoolLoong (approved):

PR #1467, opened 2024-01-02
refactor: adapt last "blocks" and clean code

Draglis:
* Red Sandstone Slab Block
* Red Sandstone Double Slab Block
  Are missing in BlockRegistry and other than that, it looks like all blocks in BlockRegistry that are not educational or experimental are implemented or adapted


src/main/java/cn/nukkit/blockentity/BlockEntity.java
```diff
@@ -24,16 +23,168 @@
  * @author MagicDroidX
  */
 @Log4j2
-public abstract class BlockEntity extends Position implements BlockEntityID {
+public abstract class BlockEntity extends Position {
+    //WARNING: DO NOT CHANGE ANY NAME HERE, OR THE CLIENT WILL CRASH
+    public static final String CHEST = "Chest";
```
CoolLoong, line 28: These move to BlockEntityID

CoolLoong (approved):

PR #1468, opened 2024-01-03
Cleanup Events and some code

Draglis:


src/main/java/cn/nukkit/event/block/BigDripleafTiltChangeEvent.java
```diff
@@ -32,6 +30,10 @@ public BigDripleafTilt getNewTilt() {
         return newTilt;
     }
 
+    public void setOldTilt(BigDripleafTilt oldTilt) {
```
CoolLoong, line 33: please remove the method

CoolLoong (changes requested):

CoolLoong (approved):

PR #1472, opened 2024-01-17
Fix #1471

### LINKED ISSUE #1471: Mud block can be broken only with shovel
# 🐞 I found a bug | 我找到个BUG

## 📝 Bug Description | BUG描述

[Mud block](https://github.com/PowerNukkitX/PowerNukkitX/blob/master/src/main/java/cn/nukkit/block/BlockMud.java) can be broken only with shovel, but [wiki](https://minecraft.fandom.com/wiki/Mud#Breaking) says it can be broken by hand.

## 🔍 Steps to Reproduce | 复现方法

Try to mine it with hand

## 💡 Possible Solution | 可能的解决方案

remove getToolTier() method

KoshakMineDEV:


AzaleeX (approved):
Super for your quick contribution ❤️
(15 min before the issue creates already finished 😅)

PR #1475, opened 2024-01-24
Convert to Gradle & Java 21

Loudbooks:


CoolLoong (changes requested):
1.I'm not going to upgrade to jdk21.
2.the .gradle folder should be excluded
3.The build logic is not equivalent and I can't merge it directly, which would break the github action.
you should extract all the dependencies to build/libs and output the source.jar and the javadoc generated by running delombok, you can see the logic in the maven/build code block.
This is not my highest priority task now, so I won't be making it, if you're interested, you can complete it.

PR #1480, opened 2024-01-29
Create real support for CTRL + C to make it safer

AzaleeX:
I've noticed an error: when you close the server with **CTRL + C**, it corrupts the `players' data`. To solve the problem, **delete all the data** files in the '`players`' folder.

### DIFF
--- a/src/main/java/cn/nukkit/utils/ServerKiller.java
+++ b/src/main/java/cn/nukkit/utils/ServerKiller.java
@@ -1,5 +1,10 @@
 package cn.nukkit.utils;
 
+import cn.nukkit.Server;
+import cn.nukkit.console.NukkitConsole;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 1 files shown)


smartcmd:
`delete all the data files in the 'players' folder.` 

?

CoolLoong:
What does it mean? I didn't understand why CTRL C would broke the data

AzaleeX:
this kills the server so does not properly save player data in player, which causes the player to crash the server every time he joins to resolve the problem until then I deleted the files of the player in question

smartcmd:
> Kills the server so does not properly **save player data in player**, which causes the player to **crash the server** every time he joins to resolve the problem until then I **deleted the files of the player** in question

You are wrong, it is because the `LOCK` file in `players` folder. Delete it

AzaleeX:
yes, that's why I **reinvented** this system which at the time of **pnx** that I had done **I did it cleaner and safer to correctly close the server by saving** it as if we executed _the command /stop_ because what I had done was just a **System.exit**

smartcmd:
> yes, that's why I **reinvented** this system which at the time of **pnx** that I had done **I did it cleaner and safer to correctly close the server by saving** it as if we executed _the command /stop_ because what I had done was just a **System.exit**

`delete all the data files in the 'players' folder.` 

Are you sure it is "cleaner and safer"?

MisledWater79:
I'm not too sure if its cleaner/safer but it does seem better, instead of just crashing the server on CTRL+C, it will properly shutdown the server.

AzaleeX:
yes I think this makes it possible to no longer just crash the server but to close it correctly

CoolLoong:
Everything works fine in a windows terminal, I'll test linux later.

CoolLoong:
linux work fine

PR #1487, opened 2024-02-04
refactor: food items

KoshakMineDEV:
All food logic was transfered to their classes


AzaleeX:
have you tested your code with a custom food item ?

KoshakMineDEV:
> have you tested your code with a custom food item ?

No, but it should work😂

src/main/java/cn/nukkit/item/ItemMilkBucket.java
```diff
@@ -1,17 +1,27 @@
 package cn.nukkit.item;
 
-public class ItemMilkBucket extends ItemBucket {
+import cn.nukkit.Player;
+
+public class ItemMilkBucket extends ItemFood {
```
CoolLoong, line 5: Milk is quite unique, and it should extends the itembucket. You can implement its logical functions uniquely

CoolLoong (changes requested):

PR #1492, opened 2024-02-06
fix: set damage to item in NBTIO#getItemHelper()

MEFRREEX:
Fixed a bug when an item was dropped or a player left the server (inventory was saved) and its meta was not saved

### DIFF
--- a/src/main/java/cn/nukkit/nbt/NBTIO.java
+++ b/src/main/java/cn/nukkit/nbt/NBTIO.java
@@ -81,7 +81,8 @@ public class NBTIO {
             BlockState blockState = getBlockStateHelper(block);
             if (blockState != null) item.setBlockUnsafe(blockState.toBlock());
         }
-        if (item.getDamage() != 0) {
+
+        if (item.getDamage() == 0 && damage != 0) {
             item.setDamage(damage);
         }
         item.setCount(amount);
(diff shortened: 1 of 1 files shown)


KoshakMineDEV (approved):
LGTM

src/main/java/cn/nukkit/nbt/NBTIO.java
```diff
@@ -81,7 +81,8 @@ public static Item getItemHelper(CompoundTag tag) {
             BlockState blockState = getBlockStateHelper(block);
             if (blockState != null) item.setBlockUnsafe(blockState.toBlock());
         }
-        if (item.getDamage() != 0) {
+
+        if (item.getDamage() == 0 && damage != 0) {
```
CoolLoong, line 85: Why only update damage to 0?
 It should be
`if (item.getDamage()!=0 && damage !=0)`

CoolLoong (changes requested):

src/main/java/cn/nukkit/nbt/NBTIO.java
MEFRREEX: If you set damage when it is not equal to 0 the damage saved in nbt will not be set when the initial damage of the item is 0, for example potions initially have damage 0, because of this if you throw it away it will turn into a bottle of water. I tested this

`if (item.getDamage() !=0 && damage != 0) {`:
https://youtu.be/W-rxPchFxzo

`if (item.getDamage() == 0 && damage != 0) {`:
https://youtu.be/9g4aq0nWrUw

CoolLoong: ok why not set the damage directly
`if damage!=0` set

PR #1493, opened 2024-02-08
refactor: removed useless command

MEFRREEX:
removed the command that sent debug information to the powernukkit server


AzaleeX (approved):
yes I understand because it's no longer used spark does what debug did

CoolLoong (approved):

PR #1496, opened 2024-02-09
World Name via the /world command

AzaleeX:
A small change to the **World Name system** changes by removing the **Dimension** in the name, so it's still displayed in the /world list as in the screenshot below

image

> [!NOTE]
> you no longer need to enter the dimension, which makes ordering easier
> ### Before
> ``` 
> /world tp "world Dim0"
> ```
> ### After
> ``` 
> /world tp world
> ```

This enhances the beauty of the **_/world list command_** and makes it easier for the user to teleport just the name of the world, much more practical.

### DIFF
--- a/src/main/java/cn/nukkit/Server.java
+++ b/src/main/java/cn/nukkit/Server.java
@@ -833,7 +833,7 @@ public class Server {
                 LevelConfig levelConfig = new LevelConfig(this.getConfig().get("level-settings.default-format", "leveldb"), generatorConfig);
                 this.generateLevel(levelFolder, levelConfig);
             }
-            this.setDefaultLevel(this.getLevelByName(levelFolder + " Dim0"));
+            this.setDefaultLevel(this.getLevelByName(levelFolder));
         }
 
         this.getTickingAreaManager().loadAllTickingArea();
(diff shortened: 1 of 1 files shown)


src/main/java/cn/nukkit/Server.java
```diff
@@ -2420,13 +2420,12 @@ public boolean loadLevel(String name) {
         }
         Map<Integer, LevelConfig.GeneratorConfig> generators = levelConfig.generators();
         for (var entry : generators.entrySet()) {
-            String levelName = name + " Dim" + entry.getKey();
-            if (this.isLevelLoaded(levelName)) {
+            if (this.isLevelLoaded(name)) {
```
CoolLoong, line 2423: Reject this change, which is broken the level store because we need the dim id as a key to distinguish between different Levels

CoolLoong (changes requested):

src/main/java/cn/nukkit/Server.java
AzaleeX: but the dimension is still there, just more in the name.

CoolLoong: There is a `Map` storing level in the `Server` class, check them!

CoolLoong: We have to distinguish the levels in this named way, because `leveldb` correspond to 3 dimensions, which corresponds to 3 `Level` instances.

CoolLoong (changes requested):

src/main/java/cn/nukkit/Server.java
AzaleeX: we can't retrieve it from config.json in the created world ?

AzaleeX: @CoolLoong

CoolLoong: yes

AzaleeX: we can't retrieve it from config.json in the created world ?

CoolLoong: Unless you're going to refactor the system, but I'm not, it's a legacy from NK and requires a lot of change to the method

AzaleeX: image

AzaleeX: what pushed me to change this is the teleportation command from world to world which needs Dim0 to be teleported.
I'll try another way

CoolLoong: It's easy to use it conveniently, just make a soft enum as a complement, I'll finish it afterward

PR #1507, opened 2024-02-14
refactor: command system

MEFRREEX:
Changes: 
1. Removed deprecated classes related to commands
2. Removed deprecated and unused methods
6. Removed deprecated methods Server#dispatchCommand and CommandMap#dispatch
3. Refactored CommandEnum class
4. Public fields replaced with getters and setters
5. Added method ParamList.getResult(int index, E defaultValue)

and some other changes


CoolLoong:
I won't accept name refactoring, I can only accept modification 6

MEFRREEX:
> I won't accept name refactoring, I can only accept modification 6

What do you mean? Changing names in CommandEnum?

CoolLoong:
I mean, you shouldn't modify these field visibility, there is no difference between get setter and public

MEFRREEX:
It doesn't affect the api in any way, but okay, I can revert those changes

CoolLoong:
It affects compatibility and many plugins need to be modified. Mostly it's a pointless change

MEFRREEX:
Okay, I'll revert this changes

MEFRREEX:
Everything's okay now?

PR #1511, opened 2024-02-15
Refactoring SetEntityLinkPacket

Dhaiven:
- Add `Builder` static class in `SetEntityLinkPacket` class to remove rededant code
- Create enum `EntityLink.Type` to replace const in `SetEntityLinkPacket`


CoolLoong:
Directly access the fields of the datapacket is the design style of PNX, and I don't want to change it

Dhaiven:
why you don't want to change it ?

CoolLoong:
Keep the usage of each packet the same, rather than each different one

KoshakMineDEV:
Just change all fields in datapackets to private and make them accessible with get and set. Direct field access is not good

Dhaiven:
> Conservez la même utilisation de chaque paquet, plutôt que de chacun d’entre eux

I agree, but for enumeration, there are packages like camera that use enumeration. However, I agree that the build class is probably useless.

CoolLoong:
Enum is useful, you can keep it

Dhaiven:
ok so i revert the builder change

Dhaiven:
It's okay, I've finally made it

CoolLoong:
You can directly remove this old method

Dhaiven:
All right

CoolLoong (approved):

PR #1516, opened 2024-02-21
perf: use kqueue when available

Blackjack200:


gradle/libs.versions.toml
```diff
@@ -11,7 +11,8 @@ com-google-code-gson-gson = "2.10.1"
 com-google-guava-guava = "33.0.0-jre"
 com-lmax-disruptor = "3.4.4"
 commons-io-commons-io = "2.15.1"
-io-netty-netty-transport-classes-epoll = "4.1.101.Final"
+io-netty-netty-transport-classes-epoll = "4.1.107.Final"
```
CoolLoong, line 14: Keep it at 101 because we need to be same with the upstream version

CoolLoong (changes requested):

PR #1529, opened 2024-03-07
Fix the issue with player kick and server shutdown where the player remains stuck on the server.

AzaleeX:
It seems like you've found an effective solution to resolve the issue of player kick and server shutdown on a Debian or Linux server. Sometimes, there can be unexpected behaviors or platform-specific bugs. The most important thing is to find a solution that works, even if you don't necessarily have a clear explanation for why it works that way. If this solution resolves the problem, that's what matters most.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -3352,6 +3352,14 @@ public class Player extends EntityHuman implements CommandSender, ChunkLoader, I
         if (!this.connected.compareAndSet(true, false) && this.closed) {
             return;
         }
+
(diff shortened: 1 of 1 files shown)


CoolLoong:
I don't understand how to reproduce the bug you say, but the `DisconnectPacket` has been sent in `BedrockSession#close`. you should say how to reproduce and I can make sure the PR is correct

AzaleeX:
> I don't understand how to reproduce the bug you say, but the `DisconnectPacket` has been sent in `BedrockSession#close`. you should say how to reproduce and I can make sure the PR is correct

this is normal you don't use pnx on linux or debian

PR #1538, opened 2024-03-21
Fix terrain is already generated

tunikakeks:
See https://discord.com/channels/944227466912870410/1218990477634961498


CoolLoong:
This is a error change, waterdogpe no longer works with pnx

PR #1559, opened 2024-04-06
Fix/soul speed

C0R0X:


src/main/java/cn/nukkit/Player.java
```diff
@@ -1044,17 +1044,18 @@ protected void handleLogicInMove(boolean invalidMotion, double distance) {
             }
 
             //处理灵魂急行附魔
+            //Handling Soul Speed Enchantment
             int soulSpeedLevel = this.getInventory().getBoots().getEnchantmentLevel(Enchantment.ID_SOUL_SPEED);
             if (soulSpeedLevel > 0) {
+                Block inBlock = this.getLevelBlock();
```
CoolLoong, line 1050: Why use this block instead of down block

C0R0X: Soul Sand is 1 pixel shorter than normal blocks. This results in downBlock.isSoulSpeedCompatible() returning false when running on just 1 layer of soul sand.

C0R0X: Added the explanation and simplified the code a bit. levelBlock.down().isSoulSpeedCompatible() check needs to stay because of soul soil having normal block height

CoolLoong (approved):

PR #1570, opened 2024-04-16
fix: nether portal

KoshakMineDEV:
Now portals are working right and nether world can be set in server.properties

### DIFF
--- a/src/main/java/cn/nukkit/Server.java
+++ b/src/main/java/cn/nukkit/Server.java
@@ -854,6 +854,21 @@ public class Server {
                 this.generateLevel(levelFolder, levelConfig);
             }
             this.setDefaultLevel(this.getLevelByName(levelFolder + " Dim0"));
+
+        }
+
+        if(this.getPropertyBoolean("allow-nether") && this.getDefaultNetherLevel() == null) {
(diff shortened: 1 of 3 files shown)


src/main/java/cn/nukkit/Server.java
```diff
@@ -854,6 +854,21 @@ private void loadLevels() {
...
+        if(this.getPropertyBoolean("allow-nether") && this.getDefaultNetherLevel() == null) {
+            String levelFolder = this.getPropertyString("level-nether-name", "nether");
+            if (levelFolder == null || levelFolder.trim().isEmpty()) {
+                log.warn("level-nether-name cannot be null, using default");
+                levelFolder = "nether";
+                this.setPropertyString("level-nether-name", levelFolder);
+            }
+
+            if (this.loadLevel(levelFolder)) {
+                this.setDefaultNetherLevel(this.getLevelByName(levelFolder + " Dim0"));;
```
CoolLoong, line 869: I think you should implement this code through plugins, because the default nether world should share common leveldb with the world, called `world dim1`

CoolLoong (changes requested):

CoolLoong (approved):

PR #1576, opened 2024-04-27
fix: ConcurrentModificationException in TagUtils

C0R0X:


src/main/java/cn/nukkit/level/updater/util/TagUtils.java
```diff
@@ -10,80 +10,55 @@
...
-        if (mutable instanceof Map) {
-            Map<String, Object> map = (Map<String, Object>) mutable;
-            CompoundTag compoundTag = new CompoundTag();
-            for (Map.Entry<String, Object> entry : map.entrySet()) {
-                compoundTag.put(entry.getKey(), toImmutable(entry.getValue()));
+        return switch (mutable) {
+            case Map<?, ?> map -> {
+                CompoundTag compoundTag = new CompoundTag();
+                map.forEach((k, v) -> {
+                    if (k instanceof String stringKey && v != null) {
```
CoolLoong, line 35: if v==null,put a endtag

CoolLoong (changes requested):

src/main/java/cn/nukkit/level/updater/util/TagUtils.java
C0R0X: fixed

CoolLoong (approved):

PR #1604, opened 2024-05-30
Fix grassspread

Buddelbubi:


src/main/java/cn/nukkit/block/BlockMobSpawner.java
```diff
@@ -70,4 +70,8 @@ public boolean canHarvestWithHand() {
         return false;
     }
 
+    @Override
+    public int getLightFilter() {
```
CoolLoong, line 74: Is the mob spawner filter light?

PR #1610, opened 2024-06-11
fix: torch recipe

givegrandpaclimb:


CoolLoong:
This recipe is automatically generated, all can't manually modify, you need to modify the export, https://github.com/CoolLoong/CodeGen

PR #1616, opened 2024-06-22
fix: recipe hopper

sleepingmag:


KoshakMineDEV:
I will later fork UpdateUtils from CoolLoong to PNX organization and back "RecipeFixGen" and I will do fixes for most of the broken recipes. We shouldn't update recipes.json manually, because we will need to do it every Minecraft update.

smartcmd:
you can get the latest recipes.json from allaymc/bedrockdata

KoshakMineDEV:
> you can get the latest recipes.json from allaymc/bedrockdata

I think we are using it already, but it has some errors like "minecraft:coal" in torch recipe instead of "minecraft:coals" and other problems which makes recipes broken

PR #1621, opened 2024-06-26
fix(Netherwarts do not grow) and update(readme)

AzaleeX:


.github/img/PowerNukkitX.png
KoshakMineDEV, line 1: I think we should make a better one logo for PNX. Maybe later I will make a new identity for PNX. I suggest to remove this logo

KoshakMineDEV (changes requested):
remove logo

KoshakMineDEV (approved):

.github/img/PowerNukkitX.png
AzaleeX: I think you're right, the logo isn't very clean.

PR #1623, opened 2024-06-26
style(update style console):

AzaleeX:
remove the space at [INFO ], [WARN ] so that it gives [INFO], [WARN], for all messages of this type in the console it's cleaner and treads [main] change color (red)


AzaleeX:
image

PR #1626, opened 2024-06-26
fix: build

KoshakMineDEV:


AzaleeX (approved):
I just deleted the getName() function, which is now useless because the name is found by itself

PR #1634, opened 2024-06-29
Pnx action

AzaleeX:


AzaleeX:
try to host the pnx package on the github repo in the package tab

PR #1638, opened 2024-07-09
Codecov update config

AzaleeX:


AzaleeX:
do not accept immediately

smartcmd:
Why remove comments?

AzaleeX:
because I'm done

smartcmd:
> because I'm done

I may not have expressed myself clearly. I would like to ask why the code comments were deleted.

AzaleeX:
I can put them back if I really need to, but for example the //deop comment doesn't need to be there.

smartcmd:
You'd better add them back

smartcmd:
By the way, you need to standardize your commit message, I've never seen this format before

smartcmd:
type(range): message

PR #1641, opened 2024-07-10
feat(update libs): for powernukkitx libs update

AzaleeX:


smartcmd:
I don't agree that. Some libs should use specific version, otherwise it will cause bugs

KoshakMineDEV:
> I don't agree that. Some libs should use specific version, otherwise it will cause bugs

if everything will work fine after updating libraries - why we should stay with old one? It's better to update project to newest libs and adapt to them instead of staying with old libs

smartcmd:
> > I don't agree that. Some libs should use specific version, otherwise it will cause bugs
> 
> if everything will work fine after updating libraries - why we should stay with old one? It's better to update project to newest libs and adapt to them instead of staying with old libs

Then make sure it works fine

PR #1646, opened 2024-07-14
Fix ObjectSavannaTree

sleepingmag:
image


KoshakMineDEV (changes requested):
This is not actual fix for this. Try to make faces.length - 1 in BlockFace Plane random()

PR #1649, opened 2024-07-15
Java migration and optimize variable nukkit

AzaleeX:


KoshakMineDEV (changes requested):
I think currently we should stay with java.security. If you will read JEP 411, it's marked for removal, but currently there is no good alternatives for it in java
https://openjdk.org/jeps/411
Also this changes can bring many problems with plugins, so I think you should revert it

PR #1651, opened 2024-07-17
refactor(move ServerProperties): move all server properties functions

AzaleeX:


KoshakMineDEV (approved):
LGTM

AzaleeX:
I'm trying to unload the Server file, which is still quite full, so I can see more clearly.

PR #1652, opened 2024-07-18
chore: readme in Russian

KoshakMineDEV:
I think that PNX community should be much more international, so I think adding of README translated to other languages will be good change. I can only speak Russian and English, so I added translation in it


AzaleeX (approved):
Well done, I'll do other languages later.

PR #1663, opened 2024-07-21
add splash, lingering potions craft

sleepingmag:
adding lingering, splash potions craft
image
image


smartcmd:
This file is generated automatically. You shouldn't edit it manually

KoshakMineDEV:
> This file is generated automatically. You shouldn't edit it manually

I will check potion recipes processing later. I think we can fix it without adding new recipes

PR #1667, opened 2024-07-26
fix(chorusfruits): no longer teleports to a block but to the top like…

AzaleeX:
… minecraft vanilla


maxmwotan:
Wait up I have discovered something!
I tried out the fix and found the issue goes further and is **not inside the chorus fruit code** but inside the teleport code!

https://github.com/user-attachments/assets/4cfdcd54-604a-4e34-9ad6-840ccf0f58ef

PR #1672, opened 2024-07-31
Fix: FloatingTextParticle

tunikakeks:
After change to Skin class (64 to 32) the old value did no longer work


maxmwotan:
I see that's the same value as `SINGLE_SKIN_SIZE` in Skin, sorry for missing that

AzaleeX (approved):

Issue #1671, opened 2024-07-30
[Bug]: Teleport behaves unusually for round coordinates


maxmwotan:
### Describe the bug

I previously reported the chorusfruit teleport behaving oddly, after some testing I found out this affects the normal teleport command aswell.
The issue is actually with teleporting to round coordinates (eg y=80 will put the player at the maximum limit of y=79)
[Pullrequest 1667](https://github.com/PowerNukkitX/PowerNukkitX/pull/1667)

### To Reproduce

`/tp ~ ~ ~` works correctly
the y-Coord is `84` in the example, but `/tp ~ 84 ~` has the same issue as chorus fruit

https://github.com/user-attachments/assets/4cfdcd54-604a-4e34-9ad6-840ccf0f58ef

### Expected behavior

on block, not inside

maxmwotan:
The offset of standing on a block is exactly `0,000009894371033` or `9,894371033e-6`
therefore teleporting to `y=84` will actually arrive at the client as `( 84 - 9.894371033e-6 )` meaning `83,999990105628967`
Which will lead to the player being put inside the block below.

Fixing the symptoms should be easy, but I wonder where this issue comes from...

maxmwotan:
This spooky number actually fixes all the related issues 😕
![QCuZ1KVlIN](https://github.com/user-attachments/assets/031882d1-208a-4df5-aea8-4a957dfb767f)

https://github.com/user-attachments/assets/c80f2274-0eaf-4df7-a6c0-3ba9b4f425ef

givegrandpaclimb:
I think the issue is not the height of the code itself, maybe that the client hasn't received the chunk data when teleporting, causing it to think it's air and then fall?

Buddelbubi:
> I think the issue is not the height of the code itself, maybe that the client hasn't received the chunk data when teleporting, causing it to think it's air and then fall?

Thats not the case. The chunks are sent before the location is changed

PR #1674, opened 2024-08-08
fix(teleport): problem with the teleport that teleports us into the block

AzaleeX:


AzaleeX:
https://github.com/PowerNukkitX/PowerNukkitX/issues/1671

CoolLoong:
you need to round the coordinates of the block as a judgment instead of simply plus it directly

PR #1676, opened 2024-08-24
Fix crafting

Buddelbubi:
Fixed crafting for shift clicking


Buddelbubi:
Wait. Missed something

Buddelbubi:
Nvm. I had and old jar.. Frick you cloudnet xd
Works fine

PR #1768, opened 2025-01-01
Enhancement: Add enableCommandBlocks setting

PleaseInsertNameHere:
This PR adds a new setting to the nukkit.yml to enable or disable the usage of command blocks server wide.

If set to true, gamerules are still able to regulate command blocks.
If set to false, gamerules do not matter.

### DIFF
--- a/src/main/java/cn/nukkit/config/ServerSettings.java
+++ b/src/main/java/cn/nukkit/config/ServerSettings.java
@@ -35,6 +35,9 @@ public final class ServerSettings extends OkaeriConfig {
     @Comment("nukkit.server.settings.playersettings")
     @CustomKey("player-settings")
     private PlayerSettings playerSettings = new PlayerSettings();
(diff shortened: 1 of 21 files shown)


src/main/java/cn/nukkit/config/ServerSettings.java
```diff
@@ -62,6 +62,8 @@ public static class BaseSettings extends OkaeriConfig {
         int autosave = 6000;
         @Comment("nukkit.server.settings.baseSettings.saveUnknownBlock")
         boolean saveUnknownBlock = true;
+        @Comment("nukkit.server.settings.baseSettings.enableCommandBlocks")
```
CoolLoong, line 65: you should create a new group of settings to control the gameplay and create the corresponding language translations

CoolLoong (changes requested):

CoolLoong (approved):

PR #1781, opened 2025-01-12
Add void & redstone ready generator

tunikakeks:


Buddelbubi:
This is nothing that should be added to PNX. Please use a plugin for that.

PR #1785, opened 2025-01-13
Fix: Snow Layers not dropping snowballs when destroyed using Torches

PleaseInsertNameHere:


Buddelbubi:
Do you maybe want to overwrite the fallable item in snow layer?

PR #1796, opened 2025-01-22
refactor: Improve food and hunger for player

AzaleeX:
I'm trying to make the file cleaner as this update is a bit dated


AzaleeX:
I don't understand all this breaking the food system

PR #1810, opened 2025-02-05
Add all possible PowerNukkitX startup scripts

AzaleeX:
This update includes all possible startup **scripts** for **PowerNukkitX** by fetching them from **[PowerNukkitX/scripts](https://github.com/PowerNukkitX/scripts)**. This improves flexibility and simplifies deployment.

> [!IMPORTANT]  
> It might also be a good idea to remove the following files, as they may not be useful:  
> - `libs.tar.gz.md5`  
> - `libs.tar.gz.sha256`  
> - `powernukkitx.jar.sha256`
> it can just be put in a zip or something else


smartcmd:
Why you use wget

AzaleeX:
> Why you use wget

to download them

smartcmd:
> > Why you use wget
> 
> to download them

??? They can be touched directly, why you download them instead

AzaleeX:
> They can be touched directly, why you download them instead
Put it directly in the repo

smartcmd:
move all scripts to scripts/***

smartcmd (approved):

Buddelbubi:
@AzaleeX this caused the github actions to fail all the time

PR #1816, opened 2025-02-11
Fix: doubled ci entries + ce entries

tunikakeks:
ci = creative inv
ce = custom item


Buddelbubi:
Will beautify that tomorrow

PR #1837, opened 2025-02-19
Fix: isSupportValid for plant type blocks

OmniacDev:
some of them were missing valid blocks. This should definitely be refactored later to use a generic Growable block tag or something.


OmniacDev:
nevermind, this isn't correct

PR #1838, opened 2025-02-19
Fix: Use BlockTags for isSupportValid instead of Ids

OmniacDev:


OmniacDev:
> Could you please use BlockTags.DIRT instead of "dirt"?

Yep, done

PR #1843, opened 2025-02-20
refactor: rename Nukkit references to PowerNukkitX across the codebase

AzaleeX:


smartcmd (changes requested):
Rejected

OmniacDev:
reasons?

smartcmd:
WE DO NOT NEED THESE MEANINGLESS CHANGES

OmniacDev:
I wouldn't call it meaningless. PowerNukkitX is so far from Nukkit ATP, if anything this makes stuff less ambiguous

smartcmd:
You are not allowed to remove the original author information

OmniacDev:
according to what? the GNU General Public License v3.0 allows it as long as there is valid attribution

CoolLoong:
This change makes no sense and will only make plugin migration more difficult(more changes,more difficult).

tunikakeks:
Which plugin even uses Nukkit.java

AzaleeX:
it's true no one uses Nukkit.java moreover the mention of Nukkit is always present so your justification is not correct

smartcmd:
Let's @Buddelbubi decide.

PR #1845, opened 2025-02-20
Fix: BeaconInventory slot -> ContainerSlotType 27 does not exist

staticius:
This PR fixes an issue where placing an ore item in the Beacon inventory slot would cause a ContainerSlotType 27 does not exist error. The problem occurred due to an incorrect slot mapping, which has now been corrected.


Buddelbubi:
Please check why the client sends this slot type

PR #1928, opened 2025-03-23
Feat/1.21.70

Buddelbubi:


AzaleeX (approved):
This seems correct, but for the next time I think instead of putting `nametag.getString(‘variant’)` create an interface with all the possible tags.

Buddelbubi:
> This seems correct, but for the next time I think instead of putting `nametag.getString(‘variant’)` create an interface with all the possible tags.

There is [ClimateVariant](https://github.com/PowerNukkitX/PowerNukkitX/blob/0ec20ee72f50bc73b9bd15cc529c16180f7257b6/src/main/java/cn/nukkit/entity/ClimateVariant.java#L45)
The nametag is in place so when the entity leaves the biome, and the server reboots, it keeps the variant

src/main/java/cn/nukkit/entity/passive/EntityChicken.java
```diff
@@ -126,18 +133,18 @@ public Item[] getDrops() {
         return new Item[]{Item.get(((this.isOnFire()) ? Item.COOKED_CHICKEN : Item.CHICKEN)), Item.get(Item.FEATHER)};
     }
 
-    
-
     @Override
     protected void initEntity() {
         this.setMaxHealth(4);
         super.initEntity();
+        if(namedTag.contains("variant")) {
```
AzaleeX, line 140: yes but that's understood but look, here it's just a string of the variant name in it

Buddelbubi: We do that everywhere like that

AzaleeX: Yes

PR #1953, opened 2025-04-09
Fix NullPointerException in GamemodeCommand by filtering out null pla…

guneyilmaz0:
Added a stream filter to remove null values from the players list.


Buddelbubi:
Thank you for your contribution

PR #1988, opened 2025-05-20
Entity properties enhancement and feature parity with BDS

sircurse:
Changed the order in the constructor to be less confusing, it now follows the same order as the BDS entity properties.

Ex on BDS:
`"identifier:propertyname":        { "type": "int",  "client_sync": true,  "default": 0,   "range": [ 0, 1000653665 ] }`

**How it would be used on BDS with the change:**
`new IntEntityProperty("identifier:propertyname", 0, 0, 1000653665 ) // PROPERTYID / DEFAULT VALUE / MIN VALUE / MAX VALUE`

**You can optionally enable clientSync, so the updated info is sent to the client:**
`new IntEntityProperty("identifier:propertyname", 0, 0, 1000653665, true) // PROPERTYID / DEFAULT VALUE / MIN VALUE / MAX VALUE / CLIENT SYNC`
If clientSync is not defined it defaults to false.

**Obs: It might break some plugins that rely on entity properties, the owners would need to update their properties following the order, but it makes the code easier to read and understand, making the PNX more friendly to attract more devs from BDS API.**

**Automatic syncing to client**
When calling setIntEntityProperty, setFloatEntityProperty, setBooleanEntityProperty, or setEnumEntityProperty, the data is now automatically sent to viewers only if the property is marked with clientSync = true.

**Proper value validation**
Properties like int and float now check if the value is within the defined min/max range before accepting it.

Added **getTypedEntityProperty**
A new method getTypedEntityProperty helps find property definitions more easily, and the setXEntityProperty methods are now more consistent and easier to maintain.

**Made entity properties persistent across reloads**

**This PR aligns the entity property behavior with Bedrock Dedicated Server (BDS) logic. The following adjustments have been made:**
- Entity properties are now saved only under the "properties" compound in the NBT, and only if they were explicitly set at runtime.
- Default values are no longer redundantly stored; if a property was not changed, it is omitted from the NBT and will be re-initialized from its default at load time.
- Property definitions (such as type, range, clientSync flag, and default values) are no longer serialized or persisted — they are now defined and registered statically during runtime via EntityProperty.register(...).
- During player spawn, only properties marked as clientSync = true are sent to the client via SyncEntityPropertyPacket, matching vanilla BDS behavior.
- This avoids index mismatches and improves compatibility with the Bedrock client expectation.
- Custom entities now register properties automatically at register of the entity class, no need to call EntityProperty.register anymore neither EntityProperty.buildEntityProperty as the server handle this on startup.


Buddelbubi:
Please fix in EntityProperty#init() line 42/43

sircurse:
> Please fix in EntityProperty#init() line 42/43

Done, also did another enhancements like clientSync feature and now the setEntityProperties will send packets automatically if clientSync is set as true, so no need to send packets every time you update it on the entity.

PR #1998, opened 2025-06-04
Added support for block_placer, updated custom item definitions to match latest BDS format

sircurse:
This PR introduces full support for the minecraft:block_placer component in CustomItemDefinition. This enables items to place blocks when used, but also serves a secondary purpose: rendering the item using the block’s geometry. This makes it possible to display items in 3D (e.g., when dropped on the ground, placed in frames, holding on hand, etc.).

Changes included:
- blockPlacer(String blockId, String... useOn) method added to SimpleBuilder, fully generating the minecraft:block_placer tag in the correct format.
- Automatically places minecraft:icon inside item_properties if block_placer is not used.
- Added .makePersistent(true) method to control should_despawn flag. (for future usage)
- Introduced support for minecraft:durability.damage_chance, reading both min and max values from the item.
- Added shoulDamageArmor check for armors using the item values getDamageChanceMin and  getDamageChanceMax.
- Tags are now written in both item_tags and minecraft:tags sections to ensure compatibility with Bedrock behavior. 
- Added cooldown and use_duration as part of simple builder by default.
- Synchronized CustomItemDefinition NBT structure with the latest format used by Bedrock Dedicated Server (BDS).

These changes make the item JSON generation fully compliant with modern BDS standards and improve flexibility for both visual and functional customization.

usage of blockPlacer example:
public CustomItemDefinition getDefinition() {
        return CustomItemDefinition
                .armorBuilder(this)
                .name("Brass Helmet")
                .creativeCategory(CreativeCategory.EQUIPMENT)
                .creativeGroup("minecraft:itemGroup.name.helmet")
                //.texture("brass_helmet") // texture is not needed if blockPlacer is set
                .blockPlacer("awp:brass_helmet_icon", "awp:no_spawn_block") // first parameter is the block you want to use as geometry, the second and so on is an array of blocks that this item can be used, if you only want it to use the 3D renderer just place any fake block that will not be used in the world (but if the block doesn't exist you will receive warnings).
                .tag("awp:inner", "awp:armor", "awp:helmet")
                .allowOffHand(false)
                .handEquipped(false)
                .foil(false)
                .makePersistent(false)
                .build();
    }
    
    
image

### DIFF
--- a/src/main/java/cn/nukkit/entity/EntityHumanType.java
+++ b/src/main/java/cn/nukkit/entity/EntityHumanType.java
@@ -177,10 +177,6 @@ public abstract class EntityHumanType extends EntityCreature implements IHuman {
                 event.getCause() != DamageCause.FIRE_TICK &&
                 event.getCause() != DamageCause.FALL) { // No armor damage
 
-            if (armor.isUnbreakable() || armor.getMaxDurability() < 0) {
-                return armor;
-            }
-
             if (shouldDamageArmor(armor)) {
                 if (armor instanceof ItemShield) {
                     armor.setDamage(armor.getDamage() + (event.getDamage() >= 3 ? (int) event.getDamage() + 1 : 0));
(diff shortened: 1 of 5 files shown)


Buddelbubi:
Is this completed?

PR #2000, opened 2025-06-07
fix warn sent invalid action id 5 when PlayerBedEnterEvent

pixelwhiz:


pixelwhiz (commented):
Screenshot_2025-05-17_202152

fix warn

PR #2041, opened 2025-07-19
Feat/easy command unregister

staticius:
i heard that some developers think Server#getCommandMap().getCommand(name).unregister(); is too long


staticius:
Server.getCommandMap().unregister(); is more useful and understandable, don't you think?

AzaleeX:
mhh you'll have to check with the other members

tunikakeks (approved):

src/main/java/cn/nukkit/command/CommandMap.java
```diff
@@ -63,4 +63,16 @@ public interface CommandMap {
      */
     Command getCommand(String name);
 
+    /***
+     *
+     * @param command unregistering command name
+     */
+    void unregisterCommand(String command);
```
PleaseInsertNameHere: To match with the `register` method, maybe rename this to just `unregister`?

src/main/java/cn/nukkit/command/CommandMap.java
```diff
@@ -63,4 +63,16 @@ public interface CommandMap {
...
+
+    /***
+     *
+     * @param commands list of names of commands to be unregistering
+     */
+    void unregisterCommands(List<String> commands);
```
PleaseInsertNameHere: We could use varargs here (`String... commands`) and only need one method then

src/main/java/cn/nukkit/command/SimpleCommandMap.java
```diff
@@ -363,6 +364,19 @@ public Command getCommand(String name) {
         return null;
     }
 
+    @Override
+    public void unregisterCommand(String name) {
+        Command unRegistering = getCommand(name);
```
PleaseInsertNameHere: Just code style - but please rename `unRegistering` to `command` to match previous usages

AzaleeX (approved):

PR #2042, opened 2025-07-20
fix&enhancements/Custom Blocks Server Sync & Custom Blocks Builder Enhancement

sircurse:
**This PR brings a batch of much-needed improvements and fixes to the custom block system and entity behavior. Here’s a quick rundown:**

**Block definition caching:**

> Custom block definitions are now cached when registered, so the server isn’t rebuilding them on every event. This cuts down CPU/memory use and keeps things running smoothly, even with lots of custom blocks.

**Better collision & entity movement:**

> Collision boxes for custom and vanilla blocks are now always in sync server-side even with permutations or properties changes, so mobs won’t “float” above blocks with unusual shapes, and pathfinding/jumping is much more natural.
> Entity AI now jumps only when it makes sense (no more mobs endlessly bouncing at walls).

**Water culling fix:**

> Water next to custom blocks with lower height now culls properly, fixing some rendering issues.

**Support for alpha_test_single_sided:**

> Added this material instance for custom blocks.

**Automatic rotation for collision boxes:**

> Custom block collision is now properly rotated if set on the builder, matching visuals and physics.

**Easier builder API:**

> Standard features like mining/explosion resistance, friction, light, and map color—if set on the builder—are now automatically used server-side. No need for extra overrides.

**Chest/ender chest fix:**

> Chests now open correctly even with thin blocks (like custom thin blocks) above them, if there’s space.

________________________________________
Overall, these changes should make custom content more reliable, performant, and easier to use for plugin and pack creators.

Photos of Issues fixed in this PR:
image
image
image

### DIFF
--- a/src/main/java/cn/nukkit/block/Block.java
+++ b/src/main/java/cn/nukkit/block/Block.java
@@ -407,9 +407,6 @@ public abstract class Block extends Position implements Metadatable, AxisAligned
         return false;
     }
 
-    //public boolean isSolid() {
-    //    return true;
-    //}
     public boolean isSolid() {
         CustomBlockDefinition def = getCustomDefinition();
         if (def != null) {
--- a/src/main/java/cn/nukkit/block/Block.java
+++ b/src/main/java/cn/nukkit/block/Block.java
@@ -1,6 +1,8 @@
 package cn.nukkit.block;
 
 import cn.nukkit.Player;
+import cn.nukkit.block.customblock.CustomBlock;
+import cn.nukkit.block.customblock.CustomBlockDefinition;
 import cn.nukkit.block.property.type.BlockPropertyType;
 import cn.nukkit.blockentity.BlockEntity;
 import cn.nukkit.entity.Entity;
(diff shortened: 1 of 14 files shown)


src/main/java/cn/nukkit/block/Block.java
```diff
@@ -349,26 +372,109 @@ public boolean canBeReplaced() {
...
-     * 控制方块是否透明(默认为false)
+     * Check if above space is greatner than 0.5 for chests
      *
-     * @return 方块是否透明
+     * @return Can chest be opened with the above space?
+     */
+public boolean hasFreeSpaceAbove() {
+    Block above = this.up();
+
+    System.out.println("[Chest] Checking space above block at " + this.getLocation());
```
tunikakeks: Please remove debug logging

src/main/java/cn/nukkit/block/Block.java
```diff
@@ -349,26 +372,109 @@ public boolean canBeReplaced() {
...
+     *
+     * @return Is the block transparent?
      */
     public boolean isTransparent() {
         return false;
     }
 
+    //public boolean isSolid() {
+    //    return true;
+    //}
```
tunikakeks: No need to keep commented code

src/main/java/cn/nukkit/block/BlockTrapdoor.java
```diff
@@ -264,17 +245,16 @@ public boolean setOpen(Player player, boolean open) {
...
         return true;
     }
 
     public void playOpenCloseSound() {
-        if (isOpen()) {
-            playOpenSound();
-        } else {
-            playCloseSound();
-        }
+        if (isOpen()) playOpenSound(); else playCloseSound();
```
tunikakeks: Please move the else part in a new line

tunikakeks (changes requested):

tunikakeks (approved):

PR #2069, opened 2025-08-14
Added Submit button on modal forms (CustomForm) and early return on getSafeSpawn when radius = 0

sircurse:
Added Submit button on modal forms (CustomForm) and early return on getSafeSpawn when radius = 0


sircurse:
> Make submitText a variable inside CustomForm instead (see parent class Form for reference)

I did in this way so it keep the same standard format to add a custom text to submit, like this:
form.addElement(new ElementSubmit("§l§0My Custom Label!"));

If it is not added the form will only send the system standard without customization.

PR #2081, opened 2025-08-30
fix: Flags split FLAGS/FLAGS_2 and match BDS sneaking/blocking (incl. HEIGHT) recalculation

sircurse:
- Separate storage for FLAGS and FLAGS_2 in EntityDataMap, route high-bit flags (≥64) to FLAGS_2.
- Implement setBlocking: with BLOCKING on/off + one-shot TRANSITION_BLOCKING (as BDS does).
- Improved setSneaking: toggling SNEAKING, removing CAN_SHOW_NAME when sneaking, and updating HEIGHT (via recalculateBoundingBox (BDS parity)

**Enables Molang tests (query.is_blocking) to behave correctly.**

### DIFF
--- a/src/main/java/cn/nukkit/entity/Entity.java
+++ b/src/main/java/cn/nukkit/entity/Entity.java
@@ -672,7 +672,30 @@ public abstract class Entity extends Location implements Metadatable, EntityID, 
     }
 
     public void setSneaking(boolean value) {
-        this.setDataFlag(EntityFlag.SNEAKING, value);
+        EnumSet<EntityFlag> primary = this.getEntityDataMap().getOrCreateFlags();
+        boolean flagsChanged = false;
+
+        if (value ? primary.add(EntityFlag.SNEAKING) : primary.remove(EntityFlag.SNEAKING)) {
+            flagsChanged = true;
+        }
+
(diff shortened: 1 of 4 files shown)


src/main/java/cn/nukkit/entity/Entity.java
```diff
@@ -672,7 +672,30 @@ public boolean isSneaking() {
...
+        EnumSet<EntityFlag> primary = this.getEntityDataMap().getOrCreateFlags();
+        boolean flagsChanged = false;
+
+        if (value ? primary.add(EntityFlag.SNEAKING) : primary.remove(EntityFlag.SNEAKING)) {
+            flagsChanged = true;
+        }
+
+        if (primary.remove(EntityFlag.CAN_SHOW_NAME)) {
+            flagsChanged = true;
+        }
```
tunikakeks: Add `CAN_SHOW_NAME` back when?

tunikakeks (changes requested):

src/main/java/cn/nukkit/entity/Entity.java
sircurse: These flags was not being used used by players, so clients was ignoring it at all, so I removed it totally.

tunikakeks (approved):

PR #2096, opened 2025-09-18
enhancement: Custom Entity Refactor - Phase 1 (!! BREAKING API !!)

sircurse:
This refactor makes custom entity creation much simpler, mirroring the approach used for **Custom Items** and **Blocks**, by introducing a `simpleBuilder` API for `CustomEntityDefinition`. It reduces boilerplate, standardizes registration, and automatically wires up spawn eggs.

* **`simpleBuilder` for entities.** Define identifier, spawn-egg presence, and summonability with a small, fluent builder.
* **One-line registration** (same pattern as items/blocks):

  ```java
  Registries.ENTITY.registerCustomEntity(plugin, MyCustomMob.class);
  ```
* **Automatic spawn-egg creation.** If `hasSpawnEgg(true)` is set, a matching egg is created and placed in the **Mob Eggs** creative group. Eggs behave like vanilla ones.
* **No client components on eggs.** Like BDS, spawn eggs are plain items. If you want a nicer egg name, localize it via lang files, the client ignores item components for eggs. (Server-side metadata is carried via `pnx_extra_data` only.)

## Usage example

```java
public class MyCustomMob extends EntityIntelligent implements CustomEntity {
    public static final String IDENTIFIER = "namespace:mycustom_mob";

    @Override public String getOriginalName() { return "My Custom Mob"; }
    @Override public @NotNull String getIdentifier() { return IDENTIFIER; }

    public MyCustomMob(IChunk chunk, CompoundTag nbt) { super(chunk, nbt); }

    // Static definition used by the registry
    public static CustomEntityDefinition definition() {
        return CustomEntityDefinition.simpleBuilder(IDENTIFIER)
            .eid(IDENTIFIER)
            .hasSpawnEgg(true)
            .isSummonable(false)
            .build();
    }

    // Instance override remains for runtime compatibility
    @Override public CustomEntityDefinition getDefinition() { return definition(); }
}
```

## Breaking changes
This refactor **breaks existing custom entities** that don’t expose a `CustomEntityDefinition` via the new pattern.

### Migration guide
1. Add a **static** definition source to each entity class (supported options, in order of preference):
   * `public static CustomEntityDefinition definition()`
   * `public static CustomEntityDefinition getDefinition()`
2. Add a instance  of `getDefinition()` if you rely on it at runtime; the registry prefers a static definition for registration.
3. Register with:
   ```java
   Registries.ENTITY.registerCustomEntity(plugin, MyCustomMob.class);
   ```

## Notes
Next phases will introduce more nested builder methods to allow simple customization of the custom entities.

### DIFF
--- a/src/main/java/cn/nukkit/command/defaults/SummonCommand.java
+++ b/src/main/java/cn/nukkit/command/defaults/SummonCommand.java
@@ -9,6 +9,7 @@ import cn.nukkit.command.tree.ParamList;
 import cn.nukkit.command.utils.CommandLogger;
 import cn.nukkit.entity.Entity;
 import cn.nukkit.level.Position;
+import cn.nukkit.registry.EntityRegistry;
 import cn.nukkit.registry.Registries;
 
 import java.util.ArrayList;
@@ -44,6 +45,7 @@ public class SummonCommand extends VanillaCommand {
             return 0;
         }
         Integer entityId = Type.ENTITY_TYPE2ID.get(entityType);
+
         Position pos = sender.getPosition();
         if (list.hasResult(1)) {
             pos = list.getResult(1);
@@ -52,6 +54,7 @@ public class SummonCommand extends VanillaCommand {
             log.addError("commands.summon.outOfWorld").output();
             return 0;
         }
+
         String nameTag = null;
         if (list.hasResult(2)) {
             nameTag = list.getResult(2);
@@ -60,18 +63,34 @@ public class SummonCommand extends VanillaCommand {
         if (list.hasResult(3)) {
             nameTagAlwaysVisible = list.getResult(3);
         }
+
+        // ---- block non-summonable entities BEFORE creating the instance ----
+        EntityRegistry.EntityDefinition def;
+        if (entityId != null) {
+            String fullId = Registries.ENTITY.getEntityIdentifier(entityId);
+            def = Registries.ENTITY.getEntityDefinition(fullId);
+        } else {
+            def = Registries.ENTITY.getEntityDefinition(entityType);
+        }
+        if (def == null || !def.summonable()) {
+            log.addError("commands.summon.failed").output();
+            return 0;
+        }
+        // -------------------------------------------------------------------
+
         Entity entity;
         if (entityId != null) {
-            //原版生物
+            // Vanilla Entities
             entity = Entity.createEntity(entityId, pos);
         } else {
-            //自定义生物
+            // Custom Entities
             entity = Entity.createEntity(entityType, pos);
         }
         if (entity == null) {
             log.addError("commands.summon.failed").output();
             return 0;
         }
+
         if (nameTag != null) {
             entity.setNameTag(nameTag);
             entity.setNameTagAlwaysVisible(nameTagAlwaysVisible);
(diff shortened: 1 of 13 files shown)


src/main/java/cn/nukkit/command/tree/node/ItemNode.java
```diff
@@ -40,6 +40,12 @@ public void fill(String arg) {
...
         }
 
-        this.value = item;
+        if (item instanceof ItemCustomEntitySpawnEgg egg) {
+             // Resolve custom entity spawn egg
+            egg.resolveSpawnEgg(arg);
+            this.value = egg;
+        } else {
+            this.value = item;
+        }
```
PleaseInsertNameHere: Duplicate with `Item.get()`

src/main/java/cn/nukkit/command/defaults/SummonCommand.java
```diff
@@ -60,18 +63,34 @@ public int execute(CommandSender sender, String commandLabel, Map.Entry<String,
...
         }
+
+        // ---- block non-summonable entities BEFORE creating the instance ----
+        EntityRegistry.EntityDefinition def;
+        if (entityId != null) {
+            String fullId = Registries.ENTITY.getEntityIdentifier(entityId);
+            def = Registries.ENTITY.getEntityDefinition(fullId);
+        } else {
+            def = Registries.ENTITY.getEntityDefinition(entityType);
+        }
```
PleaseInsertNameHere: We can make this more readable:
```java
String fullId = (entityId != null)
        ? Registries.ENTITY.getEntityIdentifier(entityId)
        : entityType;

EntityRegistry.EntityDefinition def = Registries.ENTITY.getEntityDefinition(fullId);
```

src/main/java/cn/nukkit/item/Item.java
```diff
@@ -205,8 +205,14 @@ public static Item get(String id, int meta, int count, byte[] tags, boolean auto
...
-        } else if (autoAssignStackNetworkId) {
-            item.autoAssignStackNetworkId();
+        } else {
+            if (item instanceof ItemCustomEntitySpawnEgg egg) {
+                egg.resolveSpawnEgg(id);
+            }
+
+            if (autoAssignStackNetworkId) {
+                item.autoAssignStackNetworkId();
+            }
```
PleaseInsertNameHere: Avoid nesting

src/main/java/cn/nukkit/item/ItemCustomEntitySpawnEgg.java
```diff
@@ -0,0 +1,226 @@
...
+                .add(new DoubleTag(block.getX() + 0.5))
+                .add(new DoubleTag(target.getBoundingBox() == null ? block.getY() : target.getBoundingBox().getMaxY() + 0.0001f))
+                .add(new DoubleTag(block.getZ() + 0.5)))
+            .putList("Motion", new ListTag<DoubleTag>()
+                .add(new DoubleTag(0))
+                .add(new DoubleTag(0))
+                .add(new DoubleTag(0)))
+            .putList("Rotation", new ListTag<FloatTag>()
+                .add(new FloatTag(new Random().nextFloat() * 360))
+                .add(new FloatTag(0)));
```
PleaseInsertNameHere: Code duplication, use `Entity.getDefaultNBT()` or create a static variable
*(probably copied from `ItemSpawnEgg` so we should change it there too)*

tunikakeks (approved):

PR #2106, opened 2025-10-01
Add Javadoc auto deployment to github page

RimakiTaema:
Auto Deployment Of Javadoc to it

### DIFF
--- a/.github/workflows/JavaDoc.yml
+++ b/.github/workflows/JavaDoc.yml
@@ -3,8 +3,8 @@ name: Build & Deploy Javadoc
 on:
   push:
     branches:
-      - master   # Or your default branch
-  workflow_dispatch: # Allow manual trigger
+      - master
+  workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/JavaDoc.yml
```diff
@@ -0,0 +1,54 @@
...
+      id-token: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v5
+
+      - name: Make gradlew executable
+        run: chmod +x ./gradlew
+
+      - name: Set up JDK 17
```
Adrian646: This step doesn't match the actual java version you are using there. It would be better to change it.

Adrian646 (changes requested):

.github/workflows/JavaDoc.yml
RimakiTaema: I've changed it's just Name Mismatch

PR #2129, opened 2025-10-25
Create SECURITY.md for security policy

Psycofeu:
Added a security policy document outlining supported versions and vulnerability reporting.


tunikakeks (changes requested):
I think you're supposed to actually write something in there, not use the default template 

(Use this section to, tell them)

PR #2139, opened 2025-11-08
potential fix for correctplayermovepredictionpacket

ItsToxicV2:
This pr all it does sets the name of the rotation variable to vehicleRotation and read/write the vehicleAngularVelocity as an optional float le, the same how cloudburst/pocketmine does it


tunikakeks:
> rotation field is named correctly, that's the player's rotation, not vehicle. Otherwise this is correct

@DiddyV2 please fix

DiddyV2:
> > rotation field is named correctly, that's the player's rotation, not vehicle. Otherwise this is correct
> 
> @DiddyV2 please fix

Done

tunikakeks (approved):

PR #2143, opened 2025-11-10
Fixed boat placement on water

Buddelbubi:


src/main/java/cn/nukkit/item/ItemBoat.java
```diff
@@ -82,6 +82,10 @@ public void internalAdjust() {
                 name = "Cherry Boat";
                 this.id = ItemID.CHERRY_BOAT;
                 this.identifier = new Identifier(ItemID.CHERRY_BOAT);
+            case 9:
+                name = "Pale Oak Boat";
+                this.id = ItemID.CHERRY_BOAT;
+                this.identifier = new Identifier(ItemID.CHERRY_BOAT);
```
tunikakeks: This is not correct

tunikakeks (approved):

PR #2163, opened 2025-11-22
Rapid Hunger Drain While Flying in Survival Mode

ayrzDev:
Rapid Hunger Drain While Flying in Survival Mode


xRookieFight:
Changes in this function is unnecessary and API breaking, you can disable hunger from PlayerFood.isEnabled

xRookieFight:
And since PR #2161 you can use Player.isFlying() instead of getting adventure settings

xRookieFight:
I didn't talked about plugins, please don't use translate 😄

PR #2186, opened 2025-12-14
[Plugins] Problem with not working on the console

ayrzDev:
The issue of not being able to see plugin information on the server terminal has been fixed.

### DIFF
--- a/src/main/java/cn/nukkit/command/defaults/PluginsCommand.java
+++ b/src/main/java/cn/nukkit/command/defaults/PluginsCommand.java
@@ -12,8 +12,8 @@ import cn.nukkit.lang.TranslationContainer;
 import java.util.Map;
 
 /**
- * @author Ayrz
- * @since update 2025/12/14
+ * @author xtypr
+ * @since 2015/11/12
  */
 public class PluginsCommand extends Command implements CoreCommand {
 


src/main/java/cn/nukkit/command/defaults/PluginsCommand.java
```diff
@@ -7,12 +7,13 @@
...
 import cn.nukkit.utils.TextFormat;
+import cn.nukkit.lang.TranslationContainer;
 
 import java.util.Map;
 
 /**
- * @author xtypr
- * @since 2015/11/12
+ * @author Ayrz
+ * @since update 2025/12/14
```
xRookieFight: The `@author` tag is meant to credit the original creator of the class, and `@since` should reflect when it was first introduced, not when it was last updated.

xRookieFight (changes requested):

src/main/java/cn/nukkit/command/defaults/PluginsCommand.java
```diff
@@ -30,20 +31,25 @@ public PluginsCommand(String name) {
...
             }
             list.append(plugin.isEnabled() ? TextFormat.GREEN : TextFormat.RED);
             list.append(plugin.getDescription().getFullName());
         }
-        log.addMessage("nukkit.command.plugins.success", String.valueOf(plugins.size()), list.toString()).output();
+
+        sender.sendMessage(new TranslationContainer("nukkit.command.plugins.success", new String[]{
+            String.valueOf(plugins.size()), 
+            list.toString()
+        }));
```
xRookieFight, line 53: Varargs constructor exists. `new String[] {}` is redundant here

Buddelbubi:
I cannot reproduce this.
grafik


Just works fine. log.addMessage is the actual way how command responses should work.
Maybe try log.addSuccess instead of addMessage

src/main/java/cn/nukkit/command/defaults/PluginsCommand.java
Buddelbubi: Unless the class is a placeholder class. Like an unimplemented entity.

PR #2202, opened 2025-12-18
Add functions to send/clear debug shapes at player side

tunikakeks:


Buddelbubi:
Maybe a set of sent debug shape ids would be better? That way you could still despawn individual debug shapes.

PR #2209, opened 2025-12-19
fix: TimeUnit constant casing in Gradle build

xRookieFight:
In locales like Turkish, case conversion breaks "minutes" and results
in an invalid TimeUnit reference.

Using TimeUnit.MINUTES avoids locale-specific casing issues.
image


AzaleeX:
oupsi it's my fault

PR #2219, opened 2025-12-20
Added tooltip to CustomForm elements

xRookieFight:
# Changes
- Added tooltip to CustomForm elements
- Added addElements method
- Added basic element methods to custom form

### DIFF
--- a/src/main/java/cn/nukkit/form/element/custom/ElementDropdown.java
+++ b/src/main/java/cn/nukkit/form/element/custom/ElementDropdown.java
@@ -20,27 +20,27 @@ public class ElementDropdown extends Element implements ElementCustom {
 
     protected String text;
     protected List<String> options;
-    protected String tooltip;
     protected int defaultOption;
+    protected String tooltip;
 
     public ElementDropdown() {
-        this("", new ArrayList<>(), null, 0);
+        this("", new ArrayList<>(), 0, null);
     }
 
     public ElementDropdown(String text) {
-        this(text, new ArrayList<>(), null, 0);
+        this(text, new ArrayList<>(), 0, null);
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/cn/nukkit/form/window/CustomForm.java
```diff
@@ -48,6 +48,174 @@ public CustomForm addElement(ElementCustom element) {
...
+    }
+
+    public CustomForm toggle(String text, boolean defaultValue) {
+        return this.addToggle(text, defaultValue);
+    }
+
+    public CustomForm toggle(String text, boolean defaultValue, String tooltip) {
+        return this.addToggle(text, defaultValue, tooltip);
+    }
+
```
PleaseInsertNameHere: We do not need both add[type] and [type] as method. Please choose one.

src/main/java/cn/nukkit/form/element/custom/ElementDropdown.java
```diff
@@ -17,45 +17,73 @@
 @Accessors(chain = true, fluent = true)
 @AllArgsConstructor
 public class ElementDropdown extends Element implements ElementCustom {
+
     protected String text;
     protected List<String> options;
+    protected String tooltip;
```
PleaseInsertNameHere: Make `tooltip` the last element of the constructor to match other elements' constructors.

src/main/java/cn/nukkit/form/window/CustomForm.java
xRookieFight: What do you suggest? I think we can do [type]

xRookieFight: Discussed on discord with all core developers, we are using **add[type]**

PR #2239, opened 2025-12-29
enhancement: Added World Dynamic Properties

sircurse:
This enhancement added the ability to create world/server dynamic properties, a low cost, fast and easy method to add metadata saving to level DB files.

The data is persistent, and can be used across multiple levels.


src/main/java/cn/nukkit/Server.java
```diff
@@ -2794,10 +2802,441 @@ public static void setDefaultDynamicPropertiesGroupUUID(String uuid) {
...
+     */
+    public Server removeDynamicProperty(String key) {
+        LevelDBProvider provider = getWorldDynamicPropertiesProvider();
+        if (provider == null) return this;
+
+        CompoundTag root = provider.getWorldDynamicProperties();
+        if (root == null) return this;
+
+        if (!root.contains(DP_ROOT)) return this;
+        CompoundTag dyn = root.getCompound(DP_ROOT);
```
PleaseInsertNameHere: Format doesn't match other usages, use null check instead

src/main/java/cn/nukkit/Server.java
```diff
@@ -2794,10 +2802,441 @@ public static void setDefaultDynamicPropertiesGroupUUID(String uuid) {
...
+        list.add(new FloatTag((float) z));
+
+        LevelDBProvider provider = getWorldDynamicPropertiesProvider();
+        if (provider == null) return this;
+
+        CompoundTag g = ensureWorldDynamicPropertiesGroup(provider, DP_DEFAULT_GROUP_UUID);
+        g.putList(key, list);
+        saveWorldDynamicPropertiesGroup(provider, DP_DEFAULT_GROUP_UUID, g);
+        return this;
+    }
```
PleaseInsertNameHere, line 2983: Reuse `setVec3DynamicProperty(String, Vector3)`

PleaseInsertNameHere (changes requested):

PR #2252, opened 2026-01-05
Changed camera fade color to float from integer

xRookieFight:


src/main/java/cn/nukkit/command/defaults/CameraCommand.java
PleaseInsertNameHere: Color should be changed here, too.

xRookieFight: Oops i forgot it

PR #2253, opened 2026-01-05
Added input permission support

xRookieFight:
# Changes
- Added input permission command
- Added input permission functions to Player
- Changed lockComponentId byte buffer to unsigned var int from var int
- Renamed lockComponentId variable to lockComponentData

### DIFF
--- a/src/main/java/cn/nukkit/command/defaults/InputPermissionCommand.java
+++ b/src/main/java/cn/nukkit/command/defaults/InputPermissionCommand.java
@@ -13,6 +13,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
+/**
+ * @author xRookieFight
+ * @since 2026/05/01
+ */
 public class InputPermissionCommand extends VanillaCommand {
 
     public InputPermissionCommand(String name) {


src/main/java/cn/nukkit/command/defaults/InputPermissionCommand.java
```diff
@@ -0,0 +1,150 @@
...
+        players = players.stream().filter(Objects::nonNull).toList();
+
+        if (players.isEmpty()) {
+            log.addNoTargetMatch().output();
+            return 0;
+        }
+
+        String permission = list.get(2).get();
+
+        ClientInputLocksFlag flag = switch (permission) {
```
PleaseInsertNameHere: case sensitive

src/main/java/cn/nukkit/command/defaults/InputPermissionCommand.java
```diff
@@ -0,0 +1,150 @@
...
+                boolean enable;
+
+                if ("enabled".equalsIgnoreCase(state)) {
+                    enable = true;
+                } else if ("disabled".equalsIgnoreCase(state)) {
+                    enable = false;
+                } else {
+                    log.addError("commands.inputpermission.set.missingstate").output();
+                    return 0;
+                }
```
PleaseInsertNameHere, line 113: Make `state` lowercase directly
```java
String state = list.get(3).get().toLowerCase();
```

Use switch case with default branch

PR #2257, opened 2026-01-06
Add support for registering custom packets

AzaleeX:
# 🚀 Add Custom Packet Registration for Plugins

This PR introduces a new mechanism allowing plugins to register **custom packets with unique IDs** through the new `registerCustomPacket` method.

Custom packets are now:

- Stored in a dedicated registry  
- Properly handled in `get` and `trim` operations  
- Fully isolated from official packet IDs to avoid conflicts  

> [!NOTE] 
> **Goal:** Provide a clean and extensible way for plugins to define and use their own packet types without interfering with the core protocol.

---

## 🔌 WaterDogPE Use Case

> [!IMPORTANT]
> **Why this matters**

This feature is especially useful when communicating with proxies such as **WaterDogPE**.  
It enables plugins to **send custom packets directly to the proxy**, opening the door to advanced integrations and cross‑server features.

### DIFF
--- a/src/main/java/cn/nukkit/Server.java
+++ b/src/main/java/cn/nukkit/Server.java
@@ -333,8 +333,6 @@ public class Server {
             it.saveDefaults();
             it.load(true);
         });
-        long elapsed = System.currentTimeMillis() - launchTime;
-        log.info("Successfully loaded config in {}ms.", elapsed);
 
         if (wizardConfig != null && !config.exists()) {
             this.settings.baseSettings().language(wizardConfig.getLanguage());


xRookieFight:
This seems pretty cool but we also need the usage in docs page

src/main/java/cn/nukkit/registry/PacketRegistry.java
```diff
@@ -10,30 +10,55 @@
 
 public class PacketRegistry implements IRegistry<Integer, DataPacket, Class<? extends DataPacket>> {
     private final Int2ObjectOpenHashMap<FastConstructor<? extends DataPacket>> PACKET_POOL = new Int2ObjectOpenHashMap<>(256);
+    private final Int2ObjectOpenHashMap<FastConstructor<? extends DataPacket>> customPackets = new Int2ObjectOpenHashMap<>();
```
Buddelbubi: Capitalize

src/main/java/cn/nukkit/registry/PacketRegistry.java
```diff
@@ -10,30 +10,55 @@
...
     private static final AtomicBoolean isLoad = new AtomicBoolean(false);
 
     @Override
-    public void init() {
+    public void init(){
```
Buddelbubi: revert (add space again)

src/main/java/cn/nukkit/registry/PacketRegistry.java
Buddelbubi: CUSTOM_PACKETS

Buddelbubi: Also why do you need a second map for this? This doubles the complexity O(n)

Buddelbubi (approved):

src/main/java/cn/nukkit/Server.java
```diff
@@ -333,6 +333,8 @@ public Level remove(Object key) {
             it.saveDefaults();
             it.load(true);
         });
+        long elapsed = System.currentTimeMillis() - launchTime;
+        log.info("Successfully loaded config in {}ms.", elapsed);
```
Buddelbubi: Please make this debug

PR #2287, opened 2026-01-18
Fixed problems caused by floatingtextparticles

Buddelbubi:


src/main/java/cn/nukkit/level/particle/FloatingTextParticle.java
```diff
@@ -4,42 +4,28 @@
...
-import java.awt.*;
 import java.util.ArrayList;
-import java.util.UUID;
+import java.util.EnumSet;
 import java.util.concurrent.ThreadLocalRandom;
 
 /**
- * @author xtypr
- * @since 2015/11/21
+ * Created on 2015/11/21 by xtypr.
```
xRookieFight: I don't think removing the author tag & since tag and using placeholders is the best way.

Buddelbubi: Oh. I just copied it from nukkit.
Will undo that

PleaseInsertNameHere (approved):

PR #2292, opened 2026-01-25
Simplify language selection in setup wizard

AzaleeX:
Replaces the arrow-key navigation language menu with a simpler prompt for entering a language code directly. Removes the displayLanguageMenu method and related navigation logic, streamlining the user experience during language selection.


src/main/java/cn/nukkit/wizard/SetupWizard.java
```diff
@@ -832,3 +804,6 @@ private void handleUserInputLoop(
 
 
 }
+
+
+
```
xRookieFight: These empty lines are unnecessary

PR #2312, opened 2026-02-07
Block RakNet connections & pong during server startup/shutdown

xRookieFight:
# Changes

- Prevented RakNet pong/advertisement from being sent during startup and shutdown

- Blocked new player sessions from joining while server is not fully ready

### DIFF
--- a/src/main/java/cn/nukkit/event/block/BlockChangeEvent.java
+++ b/src/main/java/cn/nukkit/event/block/BlockChangeEvent.java
@@ -1,0 +1,40 @@
+package cn.nukkit.event.block;
+
+import cn.nukkit.block.Block;
+import cn.nukkit.event.Cancellable;
+import cn.nukkit.event.HandlerList;
+import lombok.Getter;
+
+/**
+ * Called when a block changes at a specific position in the world.
+ * <p>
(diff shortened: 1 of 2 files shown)


PleaseInsertNameHere (approved):

tunikakeks (approved):

src/main/java/cn/nukkit/network/connection/BedrockPong.java
```diff
@@ -55,10 +55,24 @@ private static String toString(String string) {
...
+     * @param network the network instance to check state
+     */
     public void update(Network network) {
+        // Defensive checks
+        if (network == null) {
+            return;
+        }
+        if (channel == null) {
+            return;
+        }
```
xRookieFight, line 71: I don't get this. Are these nullable?

AzaleeX: You never know, IDK, even if it's unlikely.

AzaleeX: It's just a safety measure, as I explained, even though normally it shouldn't be null.

PR #2316, opened 2026-02-08
Player movement improvements

### LINKED ISSUE #2307: Split PlayerMoveEvent into Separate Movement and Rotation Events
# Split PlayerMoveEvent into Separate Movement and Rotation Events

## Problem
Currently, `PlayerMoveEvent` fires extremely frequently on servers with many concurrent players, causing significant performance overhead:
- Triggers every 0.1 blocks of movement
- Triggers every 1 degree of rotation
- Forces plugins to process both movement and rotation data even when they only need one

This creates unnecessary load, especially on high-population servers where plugins may only care about actual position changes (e.g., region protection) or only rotation changes (e.g., directional mechanics).

## Proposed Solution
Refactor the current event system into three distinct events:

### 1. `PlayerPositionChangeEvent` (Position Only)
- Fires only when player position changes (≥0.1 blocks)
- Does **not** fire on pure rotation changes
- Ideal for: region protection, anti-cheat position validation, teleportation triggers

### 2. `PlayerRotateEvent` (Rotation Only)
- Fires only when player pitch/yaw changes (≥1 degree)
- Does **not** fire on position changes
- Ideal for: directional mechanics, camera tracking, looking-at detection

### 3. `PlayerMoveEvent` (Legacy/Combined) - **Deprecated**
- Maintain backward compatibility by keeping current behavior
- Mark as `@Deprecated` with migration guide
- Remove in future major version

## Configuration
Add configuration options for `ROTATION_UPDATE_THRESHOLD` and `MOVEMENT_DISTANCE_THRESHOLD` in `pnx.yml`

## Benefits
- **Performance**: Plugins subscribe only to events they need, reducing callback overhead by up to 50% for position-only or rotation-only use cases
- **Clarity**: Event purpose is immediately clear from the name
- **Backward Compatible**: Existing plugins continue working during transition period
- **Scalability**: Better handles high-player-count scenarios

## Implementation Notes
- Use internal event dispatch optimization to avoid double-processing when both events have listeners
- Do not call the event at all if handlers are empty
- Event coalescing: Merge movements that happen in the same tick into one with total delta
- Consider adding a configuration option to tune movement/rotation thresholds
- Benchmark with 100+ concurrent players to validate performance improvements

xRookieFight:
## Changes

* Added **movement type support** to `PlayerMoveEvent`.
* New enum: `PlayerMoveEvent.Type`

  * `POSITION_CHANGE` → Triggered when player position changes
  * `ROTATE` → Triggered when only rotation changes

* Movement pipeline now calculates:

  * position delta
  * rotation delta
    to determine event type accurately.
* Prevents ambiguity between rotation-only and real movement events.
* Improves plugin-side performance and filtering capability.

* Made movement thresholds a setting

Closes #2307


src/main/java/cn/nukkit/event/player/PlayerMoveEvent.java
```diff
@@ -4,52 +4,48 @@
...
     private Location from;
+    @Setter @Getter
     private Location to;
 
+    @Setter @Getter
     private boolean resetBlocksAround;
 
+    @Setter @Getter
+    private Type movementType;
+
```
PleaseInsertNameHere: Use Getter and Setter at class level

src/main/java/cn/nukkit/Player.java
```diff
@@ -907,7 +905,20 @@ protected void handleMovement(Location clientPos) {
...
+            PlayerMoveEvent.Type moveType;
+
+            boolean positionChanged = last.x != now.x || last.y != now.y || last.z != now.z;
+            boolean rotationChanged = last.yaw != now.yaw || last.pitch != now.pitch || last.headYaw != now.headYaw;
+
+            if (!positionChanged && rotationChanged) {
+                moveType = PlayerMoveEvent.Type.ROTATE;
+            } else {
+                moveType = PlayerMoveEvent.Type.POSITION_CHANGE;
+            }
```
PleaseInsertNameHere: This does not respect value `ALL`

xRookieFight: Oops forgot it

PleaseInsertNameHere (approved):

PR #2320, opened 2026-02-09
Item name methods change

EngincanErgunGG:
I added a new method to the Item.java class : getOriginalName()
The reason is that when an item has a custom name, there was no existing way to retrieve the item’s original (default) name. This change makes it possible to access the original name regardless of custom naming.


EngincanErgunGG:
What I'm trying to do here is, regardless of the method, I've encountered this problem even in my own plugin. If the item in the player's hand has a custom name, we can't access the item's original name. The "getOriginalName()" function can be changed, but it will definitely be necessary for future users as well. Or we should make the idConvertToName() method public.

@tunikakeks

PR #2332, opened 2026-02-15
Move SparkInstaller to wizard package

AzaleeX:
Rename SparkInstaller from cn.nukkit.utils to cn.nukkit.wizard and update references accordingly. Changed file path and package declaration, updated Server import to cn.nukkit.wizard.SparkInstaller, and refactored initSpark signature from boolean to void (removed boolean return values and final return). No other behavioral changes intended—installer still skips download if not needed and preserves existing spark.jar.


xRookieFight:
Spark installer isn't a wizard, it does the process automatically unlike our setup wizard. So storing it in `utils` package is a good way for keeping project structure clean

PR #2343, opened 2026-02-22
Protocol Migration

### LINKED ISSUE #2339: Move network and protocol libraries to Cloudburst

AzaleeX:
## Context

Currently, PowerNukkitX relies on a custom network protocol.  
This means that all packets used for communication between the server and the client are maintained independently from the standard protocol used by Nukkit.

With every Bedrock protocol update, or any internal change related to packets, we must manually adapt our implementation: updating packet IDs, adjusting binary structures, handling new fields, removing or deprecating old formats, etc.

---

### 1. High Maintenance Cost

Every change in the network layer requires specific adjustments to our custom implementation.  
This increases the cost of updates and introduces a constant risk of errors or oversights, especially during protocol version transitions.

### 2. Risk of Inconsistencies

By maintaining a separate implementation, we gradually risk diverging from the expected client behavior.  
These divergences can lead to hard-to-diagnose bugs, particularly those related to serialization/deserialization or packet field ordering.

### 3. Duplication of Effort

A significant portion of the work already implemented and maintained in Nukkit is effectively reimplemented in PowerNukkitX.  
This creates unnecessary redundancy and complicates synchronization with upstream changes.

### 4. Integration Limitations

A custom protocol reduces interoperability with the Nukkit ecosystem.  
This may impact compatibility with certain tools, extensions, or contributions originating from the Nukkit codebase.

---

## Purpose of This Pull Request

This Pull Request proposes replacing the custom protocol with the one used by Nukkit.

Concretely, this involves:

- Aligning with the packet structures defined by Nukkit  
- Using the same encoding and decoding logic  
- Reducing divergences in the network layer  
- Simplifying future synchronization with upstream  

---

## Expected Benefits

- Significant reduction in maintenance overhead during version updates  
- Improved stability and predictability of network behavior  
- Lower risk of bugs caused by implementation differences  
- Better compatibility with the Nukkit ecosystem  
- Simplified external contributions  

---

## Conclusion

This migration aims to streamline the PowerNukkitX network layer by removing a costly custom implementation in favor of strategic alignment with Nukkit’s standard protocol.

The goal is to focus development efforts on features and optimizations rather than maintaining a divergent network foundation.


xRookieFight:
Currently @Kaooot is working for a different one

PR #2348, opened 2026-02-27
Added A Persona Skin check in the skin validity method

Baynav:
Allows for Character Skins(Persona Skins) to be used when joining servers using PNX not just Classic Skins.


tunikakeks:
We just tried ingame together & it seems to work well - I guess something went wrong on your client when sending skin data

PR #2355, opened 2026-03-04
When using a water bottle on dirt, it will be mud.

Buddelbubi:


src/main/java/cn/nukkit/block/BlockDirt.java
```diff
@@ -51,6 +53,19 @@ public int getToolType() {
 
     @Override
     public boolean onActivate(@NotNull Item item, Player player, BlockFace blockFace, float fx, float fy, float fz) {
+
+        if(item.getId().equals(Item.POTION) && item.getDamage() == 0) { // Water Bottle
```
PleaseInsertNameHere: `player` might be null, add a check

PR #2358, opened 2026-03-05
Patch lots of build errors

PleaseInsertNameHere:
This PR patches errors/warnings occuring while building.

> [!NOTE]
> Some of them are left unresolved due to possible conflicts with future updates related to the entity system.

### DIFF
--- a/src/main/java/cn/nukkit/network/Network.java
+++ b/src/main/java/cn/nukkit/network/Network.java
@@ -18,7 +18,6 @@ import io.netty.bootstrap.ServerBootstrap;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.Channel;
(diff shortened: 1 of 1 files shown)


src/main/java/cn/nukkit/network/Network.java
```diff
@@ -18,6 +18,7 @@
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import io.netty.channel.Channel;
+import io.netty.channel.DefaultSelectStrategyFactory;
```
AzaleeX: Why add this without making any code changes? If it has no use, why include it? I think it should be removed.

PleaseInsertNameHere: Had a fix in mind for deprecated EventLoopGroups, which unfortunately does not work in our current implementation. Forgot to remove that, hang on a second.

tunikakeks (approved):

PR #2380, opened 2026-03-19
BDS to PNX Player Migration System Complete

Baynav:
Well that was fun. 
Any changes just ask :)

### DIFF
--- a/src/main/java/cn/nukkit/Server.java
+++ b/src/main/java/cn/nukkit/Server.java
@@ -2309,8 +2309,11 @@ public class Server {
                 return null;
             }
             Map<Integer, LevelConfig.GeneratorConfig> map = new HashMap<>();
-            //todo nether the_end overworld
-            map.put(0, new LevelConfig.GeneratorConfig("flat", System.currentTimeMillis(), false, LevelConfig.AntiXrayMode.LOW, true, DimensionEnum.OVERWORLD.getDimensionData(), Collections.emptyMap()));
+            long seed = System.currentTimeMillis();
+
+            map.put(0, new LevelConfig.GeneratorConfig("normal", seed, false, LevelConfig.AntiXrayMode.LOW, true, DimensionEnum.OVERWORLD.getDimensionData(), Collections.emptyMap()));
+            map.put(1, new LevelConfig.GeneratorConfig("nether", seed, false, LevelConfig.AntiXrayMode.LOW, true, DimensionEnum.NETHER.getDimensionData(), Collections.emptyMap()));
+            map.put(2, new LevelConfig.GeneratorConfig("the_end", seed, false, LevelConfig.AntiXrayMode.LOW, true, DimensionEnum.THE_END.getDimensionData(), Collections.emptyMap()));
             levelConfig = new LevelConfig(LevelProviderManager.getProviderName(provider), true, map);
             try {
                 config.createNewFile();


src/main/java/cn/nukkit/utils/BedrockMigrationService.java
```diff
@@ -0,0 +1,277 @@
...
+            for (int i = 0; i < armorRaw.size(); i++) {
+
+                int slot;
+                switch (i) {
+                    case 0: slot = 36; break; // helmet
+                    case 1: slot = 37; break; // chestplate / elytra
+                    case 2: slot = 38; break; // leggings
+                    case 3: slot = 39; break; // boots
+                    default: continue;
+                }
```
PleaseInsertNameHere: Use enhanced switch here:

```java
int slot = switch (i) {
    case 0 -> 36;
    ...
}
```

src/main/java/cn/nukkit/utils/BedrockMigrationService.java
```diff
@@ -0,0 +1,277 @@
...
+            }
+
+            CompoundTag gameplay = NBTIO.read(
+                    new ByteArrayInputStream(gameplayBytes),
+                    ByteOrder.LITTLE_ENDIAN
+            );
+
+            CompoundTag converted = convertBedrockNBT(gameplay, level);
+
+            server.getLogger().info("Migration Successful for " + uuid);
```
xRookieFight: We can make this in debug, not info

src/main/java/cn/nukkit/utils/BedrockMigrationService.java
```diff
@@ -0,0 +1,277 @@
...
+        pnx.putString("Level", level.getName());
+        pnx.putInt("DimensionId", tag.contains("DimensionId") ? tag.getInt("DimensionId") : 0);
+        pnx.putString("SpawnLevel", level.getName());
+        if (tag.contains("SpawnX") && tag.contains("SpawnY") && tag.contains("SpawnZ")) {
+
+            int sx = tag.getInt("SpawnX");
+            int sy = tag.getInt("SpawnY");
+            int sz = tag.getInt("SpawnZ");
+
+            // 🔥 reject invalid spawn (0,0,0 or y <= 0)
```
xRookieFight: Unnecessary comment line and emoji

src/main/java/cn/nukkit/utils/BedrockMigrationService.java
```diff
@@ -0,0 +1,313 @@
...
+        int expLevel = tag.contains("expLevel") ? tag.getInt("expLevel") : 0;
+        float expProgress = tag.contains("EXPProgress") ? tag.getFloat("EXPProgress") : 0f;
+
+        int gamemode = tag.contains("playerGameType") ? tag.getInt("playerGameType") : 0;
+        int selectedSlot = tag.contains("SelectedInventorySlot") ? tag.getInt("SelectedInventorySlot") : 0;
+
+        short air = tag.contains("Air") ? tag.getShort("Air") : 300;
+        short fire = tag.contains("Fire") ? tag.getShort("Fire") : 0;
+        float fallDistance = tag.contains("FallDistance") ? tag.getFloat("FallDistance") : 0f;
+        boolean onGround = tag.contains("OnGround") && tag.getBoolean("OnGround");
```
PleaseInsertNameHere: You can append default values to the getters, for example:
`tag.getFloat("foodSaturationLevel", 20f);`

For default values which are equal to `0`, you can just use the normal getter, it'll default to zero if the key doesn't exist.
(For example: `tag.getFloat("FallDistance")` returns 0 if unset)

Baynav (commented):
Handled @PleaseInsertNameHere requests. 
Swapped switch expression and replaced manual checks with default value getters for NBT handling.
Anything else don't hesitate!

PR #2381, opened 2026-03-19
feat: Per-Player Rate Limiting & Botnet Detection

PleaseInsertNameHere:
# Network Security: Per-Player Rate Limiting & Botnet Detection

This PR hardens the server against packet flood attacks and coordinated botnet abuse. It replaces the existing broken global rate limiter, adds per-session inbound throttling at the Netty layer, and introduces a multi-signal botnet detector that correlates activity across independent IPs.

---

## Background

The server was vulnerable to a trivial denial-of-service: an attacker could open a single connection and send ~110,000 `ModalFormResponse` packets in one server tick, stalling the main thread indefinitely. The root cause was that `BedrockSession.tick()` drained its unbounded inbound SPSC queue without any cap. Additionally, `CommandRequestProcessor` had a `RateLimiter` field on the processor singleton - shared globally across **all** players, making it functionally useless.

---

### `BedrockSession` - Netty-layer flood gate

Two independent defences are applied before a packet ever reaches the main thread:

1. **Per-session inbound `RateLimiter`** (`maxInboundPacketsPerSecond`, default 1500/s) - checked in `onPacket()` via non-blocking `tryAcquire()`. Exceeding the limit kicks the player immediately.
2. **Per-tick packet cap** (`maxPacketsPerTick`, default 500) - applied in `tick()` as a secondary backstop; excess packets are dropped and logged rather than processed.

> [!WARNING]
> The inbound queue (`PlatformDependent.newSpscQueue()`) is unbounded. Without the per-tick cap, a burst that slips past the rate limiter between ticks could still freeze the server. Both defences are required.

---

### `PacketRateLimiter` - per-player application-layer limits

**New class** `cn.nukkit.network.security.PacketRateLimiter`. One instance lives on `PlayerHandle` for the lifetime of each player's session. Covers four packet categories:

| Category | Config key | Default |
|---|---|---|
| Commands | `maxCommandsPerSecondPerPlayer` | 10/s |
| Chat | `maxChatPerSecondPerPlayer` | 2/s |
| Form responses | `maxFormResponsesPerSecondPerPlayer` | 20/s |
| Movement | `maxMovementPacketsPerSecondPerPlayer` | 40/s |

All checks use `RateLimiter.tryAcquire()` (non-blocking). Denied packets fire `PlayerHackDetectedEvent` and are silently dropped.

> [!NOTE]
> `CommandRequestProcessor` previously held a singleton `RateLimiter` - a field on a class registered once and shared by every player. That field has been removed. All per-player limiting now goes through `PlayerHandle.packetRateLimiter`.

Processors updated: `CommandRequestProcessor`, `ModalFormResponseProcessor`, `TextProcessor`, `MovePlayerProcessor`, `MoveEntityAbsoluteProcessor` (vehicles).

---

### `BotnetDetector` - distributed flood correlation

**New class** `cn.nukkit.network.security.BotnetDetector`. Tracks a 1-second sliding window of packet counts per connected session and evaluates five corroborating signals to distinguish a coordinated botnet from coincidental high traffic.

#### Signals

| # | Signal | Fires when |
|---|---|---|
| 1 | `VOLUME` | N+ sessions simultaneously exceed the packet-rate threshold |
| 2 | `TIMING` | All flood onset times fall within 2 seconds of each other |
| 3 | `PACKET_TYPE` | 70%+ of flooding sessions share the same dominant packet type (Boyer-Moore majority vote) |
| 4 | `SUBNET` | 50%+ of flooding sessions share the same /24 subnet (or /48 for IPv6) |
| 5 | `FRESH_SESSIONS` | 70%+ of flooding sessions connected less than 30 seconds ago |

> [!NOTE]
> Sessions are keyed by `InetSocketAddress` (IP **and** port), not just IP, to avoid false positives from legitimate players behind the same NAT. Subnet grouping is used as a *signal* rather than for deduplication.

#### Evaluation throttle

Signal evaluation runs at most once every **10 seconds** (`ALERT_COOLDOWN_MS`). The cooldown is applied **before** checking the minimum score - so sub-threshold evaluations also count against the cooldown window and cannot be used to bypass it via repeated ticks.

> [!WARNING]
> The 10-second cooldown means detection lags up to 10s after the attack begins. This is intentional to avoid alert storms, but operators should treat a single `BotnetReport` as a signal to investigate rather than a guarantee the attack is still ongoing.

#### Threading model

- `registerSession` / `unregisterSession` / `recordPacket` - called from Netty I/O threads; per-session `synchronized` block for window state, volatile fields for cross-thread flood-state visibility.
- `tick()` - called from the main thread once per server tick; reads volatile flood-state fields only.

---

### `BotnetReport` - immutable attack snapshot

**New class** `cn.nukkit.network.security.BotnetReport`. Produced by `BotnetDetector.tick()` when an attack is confirmed. Carries:
- `getScore()` - number of signals that fired (1–5)
- `getSignals()` - `EnumSet<Signal>` of which signals fired
- `getFloodingSessions()` - `Set<InetSocketAddress>` of flooding sockets
- `getFloodingAddresses()` - deduplicated `Set<InetAddress>` (port stripped) for use in block rules

---

### `ServerBotnetAttackEvent` - plugin hook

**New event** `cn.nukkit.event.server.ServerBotnetAttackEvent`. Fired on the main thread when a `BotnetReport` is generated. Plugins can:
- Read the report and addresses via `getReport()` / `getSuspiciousAddresses()`
- Cancel auto-blocking by calling `setAutoBlock(false)`
- Adjust the block duration via `setBlockDurationSeconds(int)`

When `isAutoBlock()` is true (default), all flooding IPs are passed to `Network.blockAddress()` for the configured duration (`botnetAutoBlockDurationSeconds`, default 60s).

---

### Configuration (`pnx.yml` → `performance.network`)

All new settings land under the existing `networkSettings` config category:

```yaml
rateLimitEnabled: true
maxInboundPacketsPerSecond: 1500
maxPacketsPerTick: 500
maxCommandsPerSecondPerPlayer: 10
maxChatPerSecondPerPlayer: 2
maxFormResponsesPerSecondPerPlayer: 20
maxMovementPacketsPerSecondPerPlayer: 40

botnetDetectionEnabled: true
botnetSuspiciousThreshold: 300
botnetMinSuspiciousIps: 3
botnetAutoBlock: true
botnetAutoBlockDurationSeconds: 60
botnetMinScore: 2
```

> [!NOTE]
> `botnetMinScore: 2` means a report fires when at least 2 signals are active - VOLUME is always present, so one corroborating signal is sufficient. Raise this value on servers with high legitimate traffic to reduce false positives.

English translations for all new keys are included in `language/eng/lang.json`.

---

## Tests

Two new test classes under `src/test/java/cn/nukkit/network/security/`:

### `BotnetDetectorTest` (21 cases)

Uses an injectable `LongSupplier clock` constructor (package-private) for fully deterministic time control - no `Thread.sleep`.

| Area | Cases |
|---|---|
| Normal traffic | `normalTrafficDoesNotFlag` |
| Session lifecycle | `sessionUnregisterClearsFloodingState`, `staleFloodingSessionsCleanedByTick` |
| Evaluation cooldown | `evaluationThrottledDuringCooldown`, `subThresholdEvalAlsoThrottlesCooldown` |
| Min-sessions gate | `fewerThanMinSessionsProducesNoReport` |
| Signal: VOLUME | `volumeSignalAlwaysPresent` |
| Signal: TIMING | `timingSignalPresentWhenOnsetsClose`, `timingSignalAbsentWhenOnsetsSpread` |
| Signal: PACKET_TYPE | `packetTypeSignalPresentWhen70PercentMatch`, `packetTypeSignalAbsentWhenTypesDiffer` |
| Signal: SUBNET | `subnetSignalPresentWhenSameSlash24`, `subnetSignalAbsentWhenDifferentSubnets` |
| Signal: FRESH_SESSIONS | `freshSessionsSignalPresentWhenAllNew`, `freshSessionsSignalAbsentWhenSessionsOld` |
| Report content | `reportContainsFloodingAddresses`, `scoreReflectsNumberOfSignalsFired` |
| Boyer-Moore | `boyerMooreCandidateCorrectForMajority` |
| isSuspicious | `isSuspiciousReturnsTrueForFloodingIp`, `isSuspiciousReturnsFalseForCleanIp` |

### `PacketRateLimiterTest` (9 cases)

Uses Mockito to mock `NetworkSettings`. Tests first-call success for all four limiters, burst exhaustion at `rate=1/s` for each, and independence between limiters.

> [!NOTE]
> The JUnit Platform Launcher (`org.junit.platform:junit-platform-launcher`) was added as `testRuntimeOnly` in `build.gradle.kts` and declared in `libs.versions.toml`. This is required by Gradle 9's test executor when `forkEvery` is set.

---

## Files Changed

| File | Change |
|---|---|
| `BedrockSession.java` | Inbound rate limiter + per-tick cap + botnet registration |
| `Network.java` | Hosts `BotnetDetector` instance; ticks it each server tick |
| `NetworkInterface.java` | Added `getBotnetDetector()` default method |
| `PlayerHandle.java` | Added `PacketRateLimiter packetRateLimiter` field |
| `NetworkSettings.java` | 12 new config fields |
| `PlayerHackDetectedEvent.java` | Added `PACKET_FLOOD`, `MODAL_SPAM` to `HackType` |
| `CommandRequestProcessor.java` | Removed broken singleton limiter; uses `PlayerHandle` |
| `ModalFormResponseProcessor.java` | Added form-response rate limit check |
| `TextProcessor.java` | Added chat rate limit check |
| `MovePlayerProcessor.java` | Added movement rate limit check |
| `MoveEntityAbsoluteProcessor.java` | Added movement rate limit check (vehicle/boat control) |
| `BotnetDetector.java` | **New** |
| `BotnetReport.java` | **New** |
| `PacketRateLimiter.java` | **New** |
| `ServerBotnetAttackEvent.java` | **New** |
| `eng/lang.json` | 15 new translation keys |
| `BotnetDetectorTest.java` | **New** - 21 unit tests |
| `PacketRateLimiterTest.java` | **New** - 9 unit tests |
| `build.gradle.kts` | `testRuntimeOnly` launcher dep |
| `libs.versions.toml` | `junit-platform-launcher` entry |

### DIFF
--- a/src/main/java/cn/nukkit/network/Network.java
+++ b/src/main/java/cn/nukkit/network/Network.java
@@ -64,7 +64,7 @@ public class Network implements NetworkInterface {
     private final Map<InetSocketAddress, BedrockSession> sessionMap = new ConcurrentHashMap<>();
     private final Map<InetAddress, LocalDateTime> blockIpMap = new ConcurrentHashMap<>();
     private final RakServerChannel channel;
-    @Getter
+    @Getter(onMethod_ = {@Override})
     private final BotnetDetector botnetDetector;
     private BedrockPong pong;
     @Getter @Setter


xRookieFight (changes requested):
There is too much settings in pnx.yml, can we decrease it?

PR #2384, opened 2026-03-21
Added end cities

Buddelbubi:


src/main/java/cn/nukkit/level/generator/populator/the_end/EndCityPopulator.java
```diff
@@ -0,0 +1,94 @@
...
+import cn.nukkit.level.generator.object.BlockManager;
+import cn.nukkit.level.generator.object.structures.EndCityPieces;
+import cn.nukkit.level.generator.populator.Populator;
+import cn.nukkit.math.BlockVector3;
+import cn.nukkit.network.protocol.types.Rotation;
+import cn.nukkit.utils.random.Xoroshiro128;
+
+public class EndCityPopulator extends Populator {
+
+    public static final String NAME = "the_end_end_city";
```
xRookieFight, line 16: I know its required to do the name like this but is there any way to fix this? Because the name looks pretty ugly. Although its not related to this PR

Buddelbubi: you mean the_end_end ?

Thats not required since its a populator. The name is only forced for GenerateFeatures.
But we do dimension_somename

xRookieFight: I mean that looks ugly atleast in end_city. Because we already know that end city is for the end. If its optional we can let it go though

src/main/java/cn/nukkit/level/generator/object/structures/EndCityPieces.java
```diff
@@ -0,0 +1,595 @@
...
+                manager.setBlockStateAt(block, BlockAir.STATE);
+            } else if (block instanceof BlockChest) {
+                chests.add(block.asBlockVector3());
+            } else if (block instanceof BlockStandingBanner) {
+                banners.add(block.asBlockVector3());
+            } else if (block instanceof BlockFrame) {
+                itemFrames.add(block.asBlockVector3());
+            } else if (block instanceof BlockBrewingStand) {
+                brewingStands.add(block.asBlockVector3());
+            }
```
PleaseInsertNameHere: Use switch case

AzaleeX: Yeah, having if statements in a loop doesn't look great in the code

src/main/java/cn/nukkit/level/generator/populator/the_end/EndCityPopulator.java
Buddelbubi: I want to keep that like it is now because of consistency

PR #2386, opened 2026-03-21
Fix default LevelConfig generation

Baynav:
When loading existing LevelDB worlds without a config.json,
PNX generated a default config using a flat overworld-only generator.

This caused Nether and End dimensions to not load for migrated
Bedrock worlds.

This change updates the default generator configuration to:
- Use "normal" for overworld as before it was defaulting migrated worlds to "flat"
- Include Nether (1) and End (2) generators

This ensures all dimensions are properly loaded on first startup
without requiring manual config edits.

Does not affect existing worlds with an existing config.json.

### DIFF
--- a/src/main/java/cn/nukkit/Server.java
+++ b/src/main/java/cn/nukkit/Server.java
@@ -2292,7 +2292,6 @@ public class Server {
                 return null;
             }
             Map<Integer, LevelConfig.GeneratorConfig> map = new HashMap<>();
-            // Auto-generate full dimension config (Required for BDS->PNX compatibility)
             long seed = System.currentTimeMillis();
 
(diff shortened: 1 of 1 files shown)


src/main/java/cn/nukkit/Server.java
```diff
@@ -2292,8 +2292,12 @@ public boolean unloadLevel(Level level, boolean forceUnload) {
                 return null;
             }
             Map<Integer, LevelConfig.GeneratorConfig> map = new HashMap<>();
-            //todo nether the_end overworld
-            map.put(0, new LevelConfig.GeneratorConfig("flat", System.currentTimeMillis(), false, LevelConfig.AntiXrayMode.LOW, true, DimensionEnum.OVERWORLD.getDimensionData(), Collections.emptyMap()));
+            // Auto-generate full dimension config (Required for BDS->PNX compatibility)
```
PleaseInsertNameHere: Remove unnecessary comment

PR #2411, opened 2026-03-30
fix: XP Bottle spawning XP Orbs twice

TraxTRL:
The `dropXp()` method has been called twice.
This fix ensures EntityXpBottle is properly closed when it collides, acting as a guard against operating on already closed XpBottle entities.


TraxTRL:
Theoretically, you could also implement the explode() functionality directly into the dropXp() method, but I think that's more a matter of personal preference

src/main/java/cn/nukkit/entity/item/EntityXpBottle.java
```diff
@@ -67,12 +67,12 @@ public boolean onUpdate(int currentTick) {
...
         if (this.age > 1200) {
             this.kill();
+            this.close();
             hasUpdate = true;
         }
 
-        if (this.isCollided) {
-            this.kill();
-            this.dropXp();
+        if (this.isCollided && !this.closed) {
```
PleaseInsertNameHere: Checking closed state twice (once here and once in `explode()`)

PleaseInsertNameHere (changes requested):

PR #2412, opened 2026-03-30
Implemented trial chambers

Buddelbubi:


src/main/java/cn/nukkit/blockentity/BlockEntityVault.java
```diff
@@ -0,0 +1,756 @@
...
+    }
+
+    public void clearPendingRewards() {
+        setItemsToEject(List.of());
+        setTotalEjectionsNeeded(0);
+        setDisplayItem(Item.AIR);
+    }
+
+    public String resolveLootTable(boolean ominous) {
+        String normalizedLootTable = normalizeResourceKey(lootTable);
```
PleaseInsertNameHere, line 349: if-statements could be combined

src/main/java/cn/nukkit/blockentity/BlockEntityTrialSpawner.java
```diff
@@ -0,0 +1,916 @@
...
+    private String ominousProjectileKind;
+
+    public BlockEntityTrialSpawner(IChunk chunk, CompoundTag nbt) {
+        super(chunk, nbt);
+        this.entityId = this.namedTag.getString(TAG_TYPE_ID, EntityID.BREEZE);
+    }
+
+    @Override
+    protected void initBlockEntity() {
+        if (!this.namedTag.contains(TAG_SPAWN_RANGE) || !(this.namedTag.get(TAG_SPAWN_RANGE) instanceof ShortTag)) {
```
PleaseInsertNameHere, line 120: This can be simplified to use the getter only as it would return null if it doesn't hold the key.

src/main/java/cn/nukkit/blockentity/BlockEntityTrialSpawner.java
```diff
@@ -0,0 +1,916 @@
...
+            this.namedTag.putShort(TAG_SPAWN_RANGE, DEFAULT_SPAWN_RANGE);
+        }
+        if (!this.namedTag.contains(TAG_REQUIRED_PLAYER_RANGE) || !(this.namedTag.get(TAG_REQUIRED_PLAYER_RANGE) instanceof ShortTag)) {
```
PleaseInsertNameHere, line 123: This can be simplified to use the getter only as it would return null if it doesn't hold the key.

src/main/java/cn/nukkit/blockentity/BlockEntityTrialSpawner.java
```diff
@@ -0,0 +1,916 @@
...
+        player.addEffect(Effect.get(EffectType.TRIAL_OMEN)
+                .setAmplifier(amplifier)
+                .setDuration(duration)
+                .setVisible(true));
+    }
+
+    private List<Player> getDetectedPlayers() {
+        List<Player> players = new ArrayList<>();
+        double maxDistanceSquared = this.requiredPlayerRange * this.requiredPlayerRange;
+        for (Entity entity : this.level.getEntities()) {
```
PleaseInsertNameHere: Please use `this.level.getNearbyEntities()` instead of iterating through every entity.

src/main/java/cn/nukkit/blockentity/BlockEntityTrialSpawner.java
```diff
@@ -0,0 +1,916 @@
...
+                        .add(new DoubleTag(spawnPos.y))
+                        .add(new DoubleTag(spawnPos.z)))
+                .putList("Motion", new ListTag<DoubleTag>()
+                        .add(new DoubleTag(motion.x))
+                        .add(new DoubleTag(motion.y))
+                        .add(new DoubleTag(motion.z)))
+                .putList("Rotation", new ListTag<FloatTag>()
+                        .add(new FloatTag(yaw))
+                        .add(new FloatTag(pitch)));
+    }
```
PleaseInsertNameHere: `Entity.getDefaultNBT()` has the same functionality.

PR #2414, opened 2026-03-31
Implemented Trail Ruins

Buddelbubi:


tunikakeks (approved):
but what are trail ruins anyway

PR #2416, opened 2026-04-01
fix: Entity Pick Request not implemented

Kaooot:
* Implemented the ActorPickRequestPacket
* Handled the ActorPickRequestPacket
* Added the new PlayerEntityPickEvent


Buddelbubi (approved):
LGTM, but is there a reason why ItemCustomEntitySpawnEgg does not extend ItemSpawnEgg?

Response from Curse:

yes, the custom entities is registered at plugin, so it must follow a different path to rebuild the entities and the IDs
also it can exist or not, as it is defined on entity to have or not a spawn egg

PR #2423, opened 2026-04-03
fix: Shared inventory not syncing

sircurse:
This PR fixes when two or more players have the same inventory opened and item contents were not being updated when changed by some of the players.


src/main/java/cn/nukkit/inventory/request/InventoryObserverSync.java
```diff
@@ -0,0 +1,36 @@
...
+
+import cn.nukkit.Player;
+import cn.nukkit.inventory.Inventory;
+
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+public final class InventoryObserverSync {
+    public InventoryObserverSync() {
+    }
```
xRookieFight: No need to use this since the only function we call in this class is a *static function*

src/main/java/cn/nukkit/network/process/processor/ItemStackRequestPacketProcessor.java
```diff
@@ -138,15 +151,71 @@ public int getPacketId() {
...
+
+            if (action instanceof TransferItemStackRequestAction transfer) {
+                source = transfer.getSource();
+                destination = transfer.getDestination();
+            } else if (action instanceof SwapAction swap) {
+                source = swap.getSource();
+                destination = swap.getDestination();
+            } else if (action instanceof DropAction drop) {
+                source = drop.getSource();
+            }
```
PleaseInsertNameHere: Use switch case

PR #2425, opened 2026-04-03
Prevent IllegalArgumentException in BlockState property application

Baynav:
Structure generation may attempt to apply properties to blocks that do not support them (e.g., AIR).

Currently this results in IllegalArgumentException: Property ... is not supported by this block

This exception occurs inside generation loops and can
1. Cause excessive CPU usage
2. Freeze or stall the server

Instead of throwing an exception, invalid property applications are safely ignored by returning the current BlockState.

Results
1. Prevents server freezes during structure generation
2. Maintains stability without affecting valid block behavior

Extra Info
This PR focuses only on stability. The underlying issue (invalid property application during generation) may require a separate fix.
Reproduced on fresh PNX server with no plugins during natural structure generation.

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -3533,7 +3533,7 @@ public class Player extends EntityHuman implements CommandSender, ChunkLoader, I
 
         saveNBT();
 
-        if (this.level != null) {
+        if (this.level != null && this.level.getProvider() != null) {
             this.namedTag.putString("Level", this.level.getName());
 
             CompoundTag achievements = new CompoundTag();
--- a/src/main/java/cn/nukkit/level/Level.java
+++ b/src/main/java/cn/nukkit/level/Level.java
@@ -652,10 +652,29 @@ public class Level implements Metadatable {
     }
 
     public void close() {
-        if (getServer().getSettings().levelSettings().levelThread() && baseTickThread.isAlive()) {
+        if (baseTickThread.isAlive()) {
             this.baseTickGameLoop.stop();
+            try {
+                this.baseTickThread.join(5000);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+            }
+            if (baseTickThread.isAlive()) {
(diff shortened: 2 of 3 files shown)


src/main/java/cn/nukkit/block/BlockStateImpl.java
```diff
@@ -123,7 +123,13 @@ public BlockState setPropertyValue(BlockProperties properties, BlockPropertyType
             } else newPropertyValues[i] = v;
         }
         if (!succeed) {
-            throw new IllegalArgumentException("Property " + propertyValue.getPropertyType() + " is not supported by this block " + this.identifier);
+            // prevent crash loop during generation
+            // Skip invalid property application (especially important for AIR)
+            if (this.identifier.equals("minecraft:air")) {
+                return this;
+            }
```
PleaseInsertNameHere: Irrelevant

src/main/java/cn/nukkit/block/BlockStateImpl.java
```diff
@@ -157,7 +163,11 @@ public BlockState setPropertyValues(BlockProperties properties, BlockPropertyTyp
                 }
             }
             errorMsgBuilder.append(" are not supported by this block ").append(this.identifier);
-            throw new IllegalArgumentException(errorMsgBuilder.toString());
+            if (this.identifier.equals("minecraft:air")) {
+                return this;
+            }
```
PleaseInsertNameHere: Irrelevant

PleaseInsertNameHere:
> > 1. This masks the root cause rather than fixing it.
> >    Silently swallowing invalid property applications means the underlying bug will continue to run undetected. A better path might be to catch and log the issue (e.g., at debug level) so we can trace the source, rather than silently returning this.
> 
> I have the log, just wasn't sure where to dump it honestly for review, this is used as a bandaid to just allow for stability with servers online at this time. If I can dump log somewhere for review, I will also be tracing it myself aswell shortly.

We do not need your logs. You should replace the error handling with a debug log instead: `log.debug(...)`

PR #2448, opened 2026-04-09
feat: Dimension API

tunikakeks:


src/main/java/cn/nukkit/block/Block.java
```diff
@@ -387,7 +387,7 @@ public int getWalkThroughExtraCost() {
     /**
      * Controls the light level of the block
      *
-     * @return Luminance Level (0 - 15)
+     * @return Luminance Dimension (0 - 15)
```
PleaseInsertNameHere, line 390: This should still be Level

src/main/java/cn/nukkit/blockentity/ICommandBlock.java
```diff
@@ -131,7 +131,7 @@ default boolean execute() {
 
     void setExecutingOnFirstTick(boolean executingOnFirstTick);
 
-    Level getLevel();
+    Dimension getLevel();
```
PleaseInsertNameHere, line 134: Getter name misleading

src/main/java/cn/nukkit/command/defaults/TimeCommand.java
```diff
@@ -129,22 +137,24 @@ public int execute(CommandSender sender, String commandLabel, Map.Entry<String,
...
                 } else if ("noon".equals(str)) {
-                    value = Level.TIME_NOON;
+                    value = Dimension.TIME_NOON;
                 } else if ("sunrise".equals(str)) {
-                    value = Level.TIME_SUNRISE;
+                    value = Dimension.TIME_SUNRISE;
                 } else if ("sunset".equals(str)) {
-                    value = Level.TIME_SUNSET;
+                    value = Dimension.TIME_SUNSET;
                 }
```
PleaseInsertNameHere, line 151: Might as well replace this with a switch-case

PleaseInsertNameHere (commented):
Make sure that variables which changed from type `Level` to `Dimension` also changed their name to `dimension` rather than `level`.

Also looks like the modified code only handles overworld by default yet.

PR #2465, opened 2026-04-20
Implemented that water can freeze

Buddelbubi:


Buddelbubi:
This has no performance impact at all

PleaseInsertNameHere:
I've re-added the testing labels. Testing is a required step in our review process. This change affects how neighboring blocks are updated, and we need to make sure it scales well under load with many players spread across the world. We'll run some tests before we move forward with merging. Please don't remove status labels that were set by reviewers.

Buddelbubi:
This has the complexity of like growing a carrot

PleaseInsertNameHere:
Regardless of complexity, testing is standard procedure for any PR - and this one touches block update propagation, which makes it especially important. The labels stay until testing is done. Please be patient, we'll get to it.

AzaleeX (approved):

tunikakeks (changes requested):
We'd like to run some tests before merging (as stated previously)

tunikakeks (approved):
Just tested & seems like there are no problems after all

PR #2480, opened 2026-04-24
Implement infinite effects and update /effect command behavior

TraxTRL:
- Added the infinite parameter to the /effect command to assigns a duration of -1 (which the client natively treats as infinite).
- Refactored the /effect <target> clear logic to support clearing specific effects
- Adjusted the addEffect() logic in Entity.java so that newly applied finite effects do not override existing infinite effects of the same amplifier, while allowing infinite effects to correctly overwrite finite ones.
- Updated the effect ticking loop in Entity.java to skip decrementing the duration if the effect is infinite (duration == -1), preventing it from ticking down to -2 and expiring.
- Added isInfinite() and setInfinite() methods to Effect.java to provide a cleaner developer API for checking and applying infinite effects without relying on the -1 value.
- Added the commands.effect.success.infinite translation key.
- Added new getIntervall() method to Effect.java to describe ticking intervall (defaults to 0 for non-ticking effects).
- Refactored canTick() to rely on new getIntervall() method.
- Added a canTick(int) overload which supports checking against a specific tick.


xRookieFight (approved):

Buddelbubi (changes requested):
A dedicated onTick(Entity entity) would be better than modifiying canTick.
Could you do that?

PleaseInsertNameHere (dismissed):

TraxTRL:
> A dedicated onTick(Entity entity) would be better than modifiying canTick. Could you do that?

I decided to replace the canTick(Entity entity) overload with canTick(int tick) instead of just the requested onTick(Entity entity) as I couldn't find a cleaner solution without adding new duplicates.
Both finite and infinite effects share the exact same mathematical check (tick % interval == 0). Without this overload, we would have to duplicate that logic in two different places or hardcode it inside onTick.
It effectively abstracts the "source of time" from the "timing logic". This allows the parameter-less canTick() to remain a clean predicate for finite durations, while onTick can leverage the same logic using the entity's lifetime for infinite effects.
Additionally, it provides a much better API for other developers. If a developer needs to predict or check a tick point (for a UI or whatever custom system) without actually applying the effect, they can now pass any temporal value (duration, lifetime or a future tick) to the effect to get an answer.
I hope this approach still fulfills your request.

Buddelbubi (approved):

PleaseInsertNameHere (approved):

PR #2492, opened 2026-04-29
fix: Sync Inverted Daylight Detector as opposite to the regular Daylight Detector

sircurse:
Sync the inverted daylight detector against the regular detector
Inverted Daylight Detector was not activating when its pair deactivate with the same light conditions but opposite.


tunikakeks (dismissed):

Buddelbubi:
Isn't getIsFullyDarkAround unused now and can be removed?

Buddelbubi (approved):

tunikakeks (approved):

PR #2493, opened 2026-04-29
enhancement: Added missing fields to Materials

sircurse:
Added missing fields to Materials
Removed the default creation of materials from custom definition, so it allow users still to use the blocks.json from RPs.
Updated Materials javadocs.


tunikakeks (dismissed):

src/main/java/cn/nukkit/block/customblock/data/Materials.java
```diff
@@ -497,7 +991,7 @@ public enum TintMethod {
                 case "DRY_FOLIAGE": return DRY_FOLIAGE;
                 case "GRASS": return GRASS;
                 case "WATER": return WATER;
-                default: return NONE;
+                default: return null;
```
PleaseInsertNameHere: Use enhanced switch

PleaseInsertNameHere (changes requested):

PleaseInsertNameHere (approved):

tunikakeks (approved):

PR #2496, opened 2026-04-30
Fix dropping locked items bypass

TraxTRL:
- Refactor player.resetInventory()
  - Add Javadoc
  - Add helper method returnItemsFromInventory(Inventory inventory) to handle inventories one by one
    - Ensure that only items without any ItemLockMode are being dropped when inventory is full
- Add temporary CursorItem tag to save cursor inventory across sessions (to avoid loss of locked items)

Note: Behavior of player.resetInventory() is usually already handled client-side through ItemStackRequestPackets (if I read the trace logs correct). However, unlike in vanilla, CraftingGrid (and CursorInventory, unless item is locked) is not persistent across sessions, so we have to return items first before player leaves the game to avoid item loss.


src/main/java/cn/nukkit/Player.java
```diff
@@ -4657,30 +4671,52 @@ public CreativeOutputInventory getCreativeOutputInventory() {
...
+            } else {
+                Item remain = remains[0];
+                if (remain.getItemLockMode() == Item.ItemLockMode.NONE) {
+                    this.dropItem(remain);
+                    inventory.clear(slot);
+                } else if (remain.getCount() != item.getCount()) {
+                    log.warn("Partially moved {} to inventory of player {}. {} remains in {} slot {}", item, this.getName(), remain.getCount(), invName, slot);
+                    inventory.setItem(slot, remain);
+                } else {
+                    log.warn("Locked item {} in {} slot {} not moved: inventory of player {} is full and dropping locked items is forbidden. The item will persist", item, invName, slot, this.getName());
```
PleaseInsertNameHere: I'm not quite sure if we should log warnings for vanilla behavior.

TraxTRL: I honestly don't quite understand why we even have this method.
The only possible reason I can see why this exists is to surpress unwanted vanilla behavior, so I thought we should warn when we couldn't surpress it.

PleaseInsertNameHere: We can turn this into a debug message, but a warning is too much imo

PleaseInsertNameHere (approved):

Buddelbubi (approved):

PR #2498, opened 2026-05-01
feat: add /coordinates command for per-player coordinates display

Wyze3306:
Add a new /coordinates [show|hide|toggle] command (alias /coords) that toggles the showCoordinates gamerule for the executing player only via GameRulesChangedPacket, without affecting the world's actual gamerule or other players. The Bedrock client renders its native HUD when the override is sent.

The command follows the convention used by other PowerNukkitX-only commands (WorldCommand, WhitelistCommand): translation keys live in the nukkit.command.* namespace and are resolved server-side, with chat output going through CommandLogger#addMessage.

Translations are added in all 19 bundled languages.


Buddelbubi:
Why does this have to be in PowerNukkitX? This isn't vanilla and can also be done in a plugin.

Wyze3306:
> Why does this have to be in PowerNukkitX? This isn't vanilla and can also be done in a plugin.

I want to understand the inclusion criteria better, because PNX already ships several non-vanilla commands directly in core (/world, /tps, /plugins, /whitelist). What distinguishes those from /coordinates ?
In addition, it's a small piece of code that requires no maintenance and is very handy for players and developers alike.

xRookieFight:
> I want to understand the inclusion criteria better, because PNX already ships several non-vanilla commands directly in core (/world, /tps, /plugins, /whitelist). What distinguishes those from /coordinates ? In addition, it's a small piece of code that requires no maintenance and is very handy for players and developers alike.

I want to say that the commands you mentioned are pretty important commands unlike this one. However you can make this a external plugin and publish it, we appreciate that. Also, we may remove the /world command and /tps command (spark does it, also /status) but that discussion would be internally in our developer team. Otherwise the commands you mentioned are pretty major commands.

tunikakeks:
> > Why does this have to be in PowerNukkitX? This isn't vanilla and can also be done in a plugin.
> 
> I want to understand the inclusion criteria better, because PNX already ships several non-vanilla commands directly in core (/world, /tps, /plugins, /whitelist). What distinguishes those from /coordinates ?
> In addition, it's a small piece of code that requires no maintenance and is very handy for players and developers alike.

The commands you mentioned are for admins to administrate the server, your /coordinates is something a plugin should add, not the server core itself

PR #2554, opened 2026-05-17
optimization

iYozemMc:


src/main/java/cn/nukkit/level/format/leveldb/LevelDBProvider.java
```diff
@@ -502,10 +503,13 @@ public void saveChunks() {
     public void saveChunks(Collection<IChunk> chunks) {
         try (WriteBatch batch = storage.createBatch()) {
             WriteBatchHelper helper = new WriteBatchHelper();
-            CompletableFuture.runAsync(() -> chunks.parallelStream().filter(IChunk::hasChanged).forEach(chunk -> {
+            List<IChunk> changed = chunks.parallelStream()
```
Buddelbubi, line 506: You basically undid the parallel serialization. This worsens performance like 100 times

src/main/java/cn/nukkit/level/Level.java
```diff
@@ -4087,7 +4091,7 @@ public void removeEntity(Entity entity) {
 
         if (entity instanceof Player p) {
             this.players.remove(entity.getId());
-            this.playerWeatherShowMap.removeInt(p.getName());
+            this.playerWeatherShowMap.remove(p);
```
Buddelbubi, line 4094: Never use players as keys! Always use name, id, uuid etc.

Buddelbubi (commented):
No offense.. But I will just close this since I feel you dont know what you're doing here...

PR #2564, opened 2026-05-21
fix: Entity properties not syncing with different types registered

sircurse:
Fix entity properties not syncing when different types are registered for same entity
For instance
`int type first`
`float type later` <- the float will not sync.


src/main/java/cn/nukkit/entity/Entity.java
```diff
@@ -6099,42 +6099,37 @@ public final boolean setEnumEntityProperty(String identifier, String value) {
...
+                schemaIndex++;
                 continue;
             }
-            if (this.shouldSyncFloatProperty(property) && (value = this.getFloatPropertyValue(property)) != null) {
-                floats.add((float) value);
+
+            if (shouldSyncIntProperty(prop)) {
+                Integer value = getIntPropertyValue(prop);
+                if (value != null) {
+                    syncData.getIntProperties().add(new org.cloudburstmc.protocol.bedrock.data.actor.IntEntityProperty(schemaIndex, value));
```
PleaseInsertNameHere, line 6116: Use imports

src/main/java/cn/nukkit/entity/Entity.java
```diff
@@ -6099,42 +6099,37 @@ public final boolean setEnumEntityProperty(String identifier, String value) {
...
+                }
+
+                schemaIndex++;
+                continue;
             }
+
+            if (shouldSyncFloatProperty(prop)) {
+                Float value = getFloatPropertyValue(prop);
+                if (value != null) {
+                    syncData.getFloatProperties().add(new org.cloudburstmc.protocol.bedrock.data.actor.FloatEntityProperty(schemaIndex, value));
```
PleaseInsertNameHere, line 6126: Use imports

PleaseInsertNameHere (changes requested):

src/main/java/cn/nukkit/entity/Entity.java
sircurse: cant use imports because we also have classes of same name.

src/main/java/cn/nukkit/entity/Entity.java
sircurse: cant use imports because we also have classes of same name.

PleaseInsertNameHere (approved):

Buddelbubi (approved):

PR #2573, opened 2026-05-25
fix: synchronize inner chunkLoaders map to prevent ConcurrentModificationException

xRookieFight:


PleaseInsertNameHere:
Already covered by #2565

PR #2585, opened 2026-05-30
fix: spear attack not working while mounted

YnXahgkiO:
## Summary

- Spear attacks (stab + lunge) were silently ignored when the player was mounted because the speed check read `player.getMovementSpeed()` (always ~0 on a mount) instead of the mount's own velocity
- The mount entity was also included in the nearby-entity search, causing self-hits on the ridden entity
- Lunge gate was too strict: `!isSprinting()` blocked all attacks while mounted even at full mount speed
- `minDot` (cone angle) was slightly relaxed while mounted (0.75 vs 0.866) to compensate for reduced aim precision on a moving mount

## Changes

**`ItemSpear.java`**
- Speed source: use `riding.getMotion().length()` when mounted, `player.getMovementSpeed()` otherwise
- Lunge gate: `!isSprinting() && riding == null` — mounted sprint check skipped
- Search loop: skip the ridden entity to avoid self-hit
- `minDot` widened to `0.75` while mounted

**`InventoryTransactionProcessor.java`**
- `onSpearStab` call: same speed-source fix (mount motion vs player speed)

## Test plan

- [ ] Stab a mob while riding a horse at full gallop → should hit
- [ ] Lunge (right-click) a mob while mounted → should trigger correctly
- [ ] Dismount and verify normal sprint-stab still works
- [ ] Verify the ridden horse/camel is never targeted by the spear

---

### DIFF
--- a/src/main/java/cn/nukkit/item/ItemSpear.java
+++ b/src/main/java/cn/nukkit/item/ItemSpear.java
@@ -48,7 +48,9 @@ public abstract class ItemSpear extends ItemTool {
 
         applyLunge(player);
 
-        if (movementSpeed < MINIMUM_SPEED || !player.isSprinting()) {
+        Entity riding = player.getRiding();
+
+        if (movementSpeed < MINIMUM_SPEED || (!player.isSprinting() && riding == null)) {
             SoundEvent missSound = getMissSound();
             if (missSound != null) {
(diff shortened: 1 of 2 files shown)


xRookieFight (approved):

xRookieFight:
Can you do the test plans and give the feedback? I did some but i'm not sure about all of your tests in the PR description.

PR #2586, opened 2026-05-30
fix: smithing table not preserving durability of upgraded items

YnXahgkiO:
## Summary

When upgrading a damaged tool at the smithing table (e.g. a half-durability diamond sword → netherite sword), the result was shown at full durability on the client. Additionally, a damaged item could fail to match its smithing recipe entirely if the equipment descriptor performed a strict damage check.

Two fixes in `handleSmithingUpgrade`:

1. **Recipe match** — when the equipment descriptor is a `DefaultDescriptor`, compare by item ID only (ignore damage). The previous `match()` call would fail for any damaged equipment since damage > 0 doesn't equal the recipe's expected damage of 0.

2. **Result durability** — after cloning the result and copying the equipment's NBT, explicitly call `setDamage` when the input was damaged. `setNbt` updates the compound tag (which includes the `Damage` entry) but does not synchronise the item's internal damage field used in the network packet, causing the client to render full durability.

## Test plan

- [x] Upgrade a damaged diamond sword to netherite — result should keep the same durability
- [x] Upgrade a full-durability diamond sword — result should show full durability (no `Damage:0` added)
- [x] Verify enchantments and other NBT are preserved after upgrade

---

### DIFF
--- a/src/main/java/org/powernukkitx/inventory/request/CraftRecipeActionProcessor.java
+++ b/src/main/java/org/powernukkitx/inventory/request/CraftRecipeActionProcessor.java
@@ -20,6 +20,7 @@ import org.powernukkitx.recipe.Recipe;
 import org.powernukkitx.recipe.SmithingTransformRecipe;
 import org.powernukkitx.recipe.UserDataShapelessRecipe;
 import org.powernukkitx.recipe.SmithingTrimRecipe;
+import org.powernukkitx.recipe.descriptor.DefaultDescriptor;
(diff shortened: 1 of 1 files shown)


tunikakeks (approved):

TraxTRL:
Hi @YnXahgkiO 👋 I have changed the target branch to `master` (since `b/migration` has been merged and will be deleted soon). Please resolve the conflicts so we can move forward with your PR. Thank you! 😊

xRookieFight (approved):

PR #2587, opened 2026-05-30
fix: skeleton bow drop chance and durability now match vanilla

YnXahgkiO:
## Summary

Skeletons were always dropping their bow at full durability with a 100% drop rate, which does not match vanilla Bedrock behaviour.

**Vanilla behaviour:**
- Bow drop chance: 8.5% base, +1% per Looting level
- Dropped bow is heavily damaged: damage value between 193 and 384 (out of 385 max), i.e. 50–99% durability consumed

**Before:** every skeleton kill guaranteed a full-durability bow.

**After:** `getDrops` now applies the correct chance and randomises the damage value on the dropped bow. Non-bow equipment (armour, other held items) is unaffected.

---

### DIFF
--- a/src/main/java/cn/nukkit/entity/mob/EntitySkeleton.java
+++ b/src/main/java/cn/nukkit/entity/mob/EntitySkeleton.java
@@ -108,7 +108,16 @@ public class EntitySkeleton extends EntityMob implements EntityWalkable, EntityS
 
         for (Item equipped : this.getEquipmentInventory().getContents().values()) {
             if (!equipped.isNull() && !equipped.hasEnchantment(Enchantment.ID_VANISHING_CURSE)) {
-                drops.add(equipped.clone());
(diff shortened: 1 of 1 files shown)


src/main/java/cn/nukkit/entity/mob/EntitySkeleton.java
```diff
@@ -108,7 +108,16 @@ public Item[] getDrops(@NotNull Item weapon) {
 
         for (Item equipped : this.getEquipmentInventory().getContents().values()) {
             if (!equipped.isNull() && !equipped.hasEnchantment(Enchantment.ID_VANISHING_CURSE)) {
-                drops.add(equipped.clone());
+                if (equipped.getId().equals(Item.BOW)) {
```
Buddelbubi, line 111: please use equipped.isBow()

Buddelbubi (changes requested):

PR #2593, opened 2026-06-07
fix: use LITTLE_ENDIAN for SyncEntityPropertyPacket NBT encoding and decoding

### LINKED ISSUE #2562: [Bug]: SyncEntityPropertyPacket encodes NBT in BIG_ENDIAN — client silently discards schema, q.property() always returns 0
### Describe the bug

SyncEntityPropertyPacket.encode() writes NBT using ByteOrder.BIG_ENDIAN. Every other network NBT packet in PNX uses LITTLE_ENDIAN. The Bedrock client cannot parse the big-endian payload and silently drops the packet. As a result, the client never registers the entity property schema for custom entities, so q.property('name') always returns 0 in RP animations and Molang controllers.

### To Reproduce

Register a custom entity with float entity properties (e.g. msd:pitch, msd:roll) using EntityProperty.register()

Set those properties in-game via setFloatEntityProperty()

In the resource pack RP entity definition, use q.property('msd:pitch') in a pre_animation script

Observe that the animation variable never changes — always 0

### Expected behavior

q.property() returns the server-set value; animations animate

### Screenshots

_No response_

### Version

latest

### Additional context

Actual: q.property() always returns 0 regardless of what the server sets.

Root cause:
[SyncEntityPropertyPacket.java line 35](https://github.com/PowerNukkitX/PowerNukkitX/blob/master/.../SyncEntityPropertyPacket.java):

// Wrong — Bedrock network NBT is little-endian
byteBuf.writeBytes(NBTIO.write(data, ByteOrder.BIG_ENDIAN, true));
Compare with every other packet (e.g. BlockEntityDataPacket.java):

byteBuf.writeBytes(NBTIO.write(namedTag, ByteOrder.LITTLE_ENDIAN, true));
Fix:

// SyncEntityPropertyPacket.java — both encode() and decode()
ByteOrder.BIG_ENDIAN  →  ByteOrder.LITTLE_ENDIAN

chungchandev:
Fixes #2562 

Verified on a real Bedrock client. Resource pack custom entity with clientSync `FloatEntityProperty` was invisible before and renders correctly (with `query.property` resolving) after the change.

### DIFF
--- a/src/main/java/cn/nukkit/network/protocol/SyncEntityPropertyPacket.java
+++ b/src/main/java/cn/nukkit/network/protocol/SyncEntityPropertyPacket.java
@@ -23,7 +23,7 @@ public class SyncEntityPropertyPacket extends DataPacket {
     @Override
     public void decode(HandleByteBuf byteBuf) {
         try (ByteBufInputStream stream = new ByteBufInputStream(byteBuf)) {
-            this.data = NBTIO.read(stream, ByteOrder.BIG_ENDIAN, true);
+            this.data = NBTIO.read(stream, ByteOrder.LITTLE_ENDIAN, true);
         } catch (Exception e) {
             log.error("", e);
         }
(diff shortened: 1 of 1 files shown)


tunikakeks:
Thank you for your contribution. However with our new release, PNX 3.0.0, this is no longer needed.

PR #2596, opened 2026-06-08
enhancement: Improved flying mob hover roaming and speed tuning

sircurse:
Updates hover roaming behavior for flying mobs to make movement more natural and easier to tune.

Changes include:
- Keep hover mobs inside their configured `hoverMin` / `hoverMax` band.
- Improve horizontal and diagonal movement during hover roaming.
- Reduce robotic vertical-only movement by limiting abrupt Y changes.
- Allow horizontal steering even while vertical motion is active.
- Change `HoverRandomRoamExecutor` speed input from absolute speed to a multiplier.
- Compute hover roam speed from `getDefaultFlyingSpeed() * speedMultiplier`.

This makes flying mobs behave closer to BDS tuning, the base flying speed comes from the entity, and each mob can fine-tune hover roaming with a multiplier.

**PS: That is not a API breaking change, but custom entities using the HoverRandomRoamExecutor must update the signature to pass the multiplier, and override the `getDefaultFlyingSpeed()`, otherwise the entity will hover in a different speed than expected.**


PleaseInsertNameHere (approved):

xRookieFight (approved):
LGTM, just need some tests. Will do it ASAP

PR #2597, opened 2026-06-14
feat(item): add utility method to Item class

iYozemMc:


PleaseInsertNameHere (commented):
We could also start using the new feature in the codebase straight away

src/main/java/cn/nukkit/item/Item.java
```diff
@@ -2638,6 +2638,18 @@ public CompoundTag getCustomItemComponent(String key) {
...
+    public static void applyDamage(Player player, ItemTool item, int amount) {
+        if (!item.canTakeDamage() || item.isUnbreakable()) return;
+
+        item.setDamage(item.getDamage() + amount);
+
+        if (item.getDamage() >= item.getMaxDurability()) {
+            player.getInventory().setItemInMainHand(Item.AIR);
+        } else {
+            player.getInventory().setItemInMainHand(item);
+        }
```
xRookieFight, line 2725: I'm not sure, break logic should happen in Item#setDamage, right?. I don't know we need a break logic function here or not.

xRookieFight (changes requested):

xRookieFight:
> We could also start using the new feature in the codebase straight away

Yeah, we can merge this right into 'b/migration'

src/main/java/cn/nukkit/item/Item.java
tunikakeks: Iirc break logic is already implemented somewhere

xRookieFight (approved):

Buddelbubi:
What is this? Item breaking is handled somewhere else Iirc.
If you want to add those, please use them into the codebase too.

PR #2600, opened 2026-06-15
fix!: stop block-entity duplication & make chunk auto-save consistent (outdated)

TraxTRL:
See #2617 for up-to-date PR

Fixes a family of block-entity corruption bugs that traced back to a single root cause: chunks could be initialized more than once concurrently, creating duplicate block entities at the same position. This surfaced as:

- a `ClassCastException` that halted an entire level's tick loop (a `BlockEntityCrafter` left orphaned at a position now occupied by a hopper, `(BlockCrafter) getLevelBlock()` on a `BlockHopper`),
- signs (and other block entities) reverting to stale content after a restart,
- generally inflated/duplicated block-entity data accumulating rapidly near players.

It also hardens the chunk auto-save path, which independently could write a torn block/block-entity snapshot.

`Chunk#initChunk` did an unsynchronized check-then-act on `isInit`. `forceLoadChunk` is reachable concurrently from several threads (the tick thread, the async chunk loader, netty login and entity AI (`asyncPrepare` → `getBlock` → `getChunk`)), all operating on the same chunk instance. Multiple threads entering `initChunk` each re-ran the body and created a full duplicate set of the chunk's (block) entities.

Separately, auto-save serialized chunks on a background thread while the main thread kept mutating them, so the block palette and the block-entity set could be captured at different instants.

### Changes

- `Chunk#initChunk` now runs exactly once per chunk via double-checked locking (`isInit` is `volatile`, body guarded by a dedicated lock)
- Chunks are now serialized on each level's tick thread (the sole mutator) a bounded number per tick (see config) and only the LevelDB write is async (single-threaded per-provider writer, drained on close so nothing is lost on shutdown). This removes the torn-snapshot race
- New `saveChunksPerTick` (default `16`) under base settings, next to `autoSave`/`autosaveDelay`. Lower spreads a save over more ticks (smoother TPS); higher finishes sooner. Read live each tick
- The chunk serializer deduplicates block entities by position, preferring the live (position-registered) entity, and warns if a duplicate is ever seen again so pre-existing on-disk corruption is collapsed on save and regressions are visible
- `CrafterInventory` no longer hard-casts `getLevelBlock()` to `BlockCrafter`; it updates redstone only when the block really is a crafter and otherwise closes the orphaned block entity (heals a block↔block-entity type mismatch instead of throwing during tick)

### API changes

- `CrafterInventory#getComperatorOutput` was renamed to `getComparatorOutput` (typo)

Note for reviewers: This is mostly made with Claude Opus 4.8, so please be extra cautious while reviewing. I have tested those changes (even in production), but I still don't 100% trust AI with level related stuff.


src/main/java/cn/nukkit/level/format/leveldb/LevelDBChunkSerializer.java
```diff
@@ -345,14 +346,35 @@ private void deserializeTileAndEntity(DB db, IChunkBuilder builder, CompoundTag
...
+        for (BlockEntity blockEntity : chunk.getBlockEntities().values()) {
+            String posKey = blockEntity.getFloorX() + ":" + blockEntity.getFloorY() + ":" + blockEntity.getFloorZ();
+            BlockEntity existing = byPosition.get(posKey);
+            if (existing == null) {
+                byPosition.put(posKey, blockEntity);
+            } else {
+                BlockEntity authoritative = chunk.getTile(blockEntity.getFloorX(), blockEntity.getFloorY(), blockEntity.getFloorZ());
+                BlockEntity keep = authoritative != null ? authoritative : blockEntity;
+                byPosition.put(posKey, keep);
+                log.warn("[block-entity] duplicate block entities at {},{},{} in chunk ({}, {}): '{}' (id {}) and '{}' (id {}); keeping '{}' (id {}). Stale/orphaned block entity skipped on save.",
```
xRookieFight:
```suggestion
                log.debug("[block-entity] Duplicate block entities at {},{},{} in chunk ({}, {}): '{}' (id {}) and '{}' (id {}); keeping '{}' (id {}). Stale/orphaned block entity skipped on save.",
```

xRookieFight (changes requested):

xRookieFight:
Haven't tested yet, @PowerNukkitX/core-dev please make the reviews with testing

TraxTRL:
This is running in production for about 7 hours now and so far it looks good.
I must highlight, I haven't tested it with level threads, so if anyone is able to, please test it with that enabled.

Still, please only approve and merge if you have ran your own tests.

src/main/java/cn/nukkit/level/format/leveldb/LevelDBChunkSerializer.java
```diff
@@ -368,6 +390,7 @@ private void serializeTileAndEntity(WriteBatch writeBatch, IChunk chunk) {
                                 chunk.getX(), chunk.getZ(), e);
                     }
                 }
+                log.debug("[block-entity] chunk ({}, {}): wrote {} block entit(y/ies)", chunk.getX(), chunk.getZ(), byPosition.size());
```
TraxTRL: This should get dropped to TRACE or removed after testing; this information is not really of use anymore

TraxTRL:
For over 24 hours in production now, looking good so far!

tunikakeks (approved):

tunikakeks:
Does this still need testing with levelthread?

TraxTRL:
> Does this still need testing with levelthread?

I haven't tested it with levelthreads yet, so probably yes.

TraxTRL:
Continuing in PR #2617 because it wouldn't let me rebase without resolving dead conflicts.

PR #2604, opened 2026-06-19
fix!: store level time as long to stop the day/night cycle freezing

TraxTRL:
This PR fixes the day/night cycle randomly freezing and "resetting every 30 seconds" on long-running worlds, plus two correctness bugs in the `/time` command.

On long-lived worlds the time of day would appear to advance, then snap back every ~30 seconds, eventually freezing entirely. Reproduced with `/time query` reading `16,790,000`,  just past 2^24 (16 777 216).

`Level.time` was a `float`. A `float` can only represent consecutive integers up to 2^24 ≈ 16.7M ticks (~9.7 days of active ticking). Past that, the spacing between representable values grows to 2+, so `this.time += 1` in `checkTime()` rounds straight back to the same value → the server clock freezes. The client free-runs the daylight cycle locally and gets snapped back to the frozen value by the periodic `SetTimePacket` (every 30s), which is the visible "reset".

The provider already persisted time as a `long`, so the value was being narrowed on load and truncated on save.

### Changes

- `Level.time`: `float` → `long`
- `getTime()` / `setTime()`: `int` → `long`
- `getDayTime()` casts the `long` modulo back to `int`; `calculateCelestialAngle` now takes a `long`
- Save path persists the full `long` instead of truncating through `int`
- `Player` `SetTimePacket` senders cast to `int` for the wire field
- `/time set <value>` and `/time set <preset>` now set only the day-time component, preserving the elapsed-day count (`setTime(getTime() - getDayTime() + value)`) instead of overwriting the cumulative clock. Previously `/time set day` rewound game time toward 0
- All `/time` subcommands (`set`/`add`/`start`/`stop`/`query`) now act on the command sender's level only, instead of looping over every loaded level and modifying all worlds at once
- Minor refactor (switch expression for presets, `final` locals)

### API changes
- `Level#getTime()` now returns `long` (was `int`)
- `Level#setTime(long)` (was `setTime(int)`)
- `Level#calculateCelestialAngle(long, float)` (was `int`)

### DIFF
--- a/src/main/java/cn/nukkit/Player.java
+++ b/src/main/java/cn/nukkit/Player.java
@@ -697,7 +697,7 @@ public class Player extends EntityHuman implements CommandSender, ChunkLoader, I
         this.setEnableClientCommand(true);
 
         final SetTimePacket setTimePacket = new SetTimePacket();
-        setTimePacket.setTime(this.level.getTime());
+        setTimePacket.setTime((int) this.level.getTime());
         this.sendPacket(setTimePacket);
 
         this.noDamageTicks = 60;
@@ -4971,8 +4971,8 @@ public class Player extends EntityHuman implements CommandSender, ChunkLoader, I
                 playerChunkManager.getUsedChunks().clear();
             }
 
-            SetTimePacket setTime = new SetTimePacket();
-            setTime.setTime(level.getTime());
+            final SetTimePacket setTime = new SetTimePacket();
+            setTime.setTime((int) level.getTime());
             this.sendPacket(setTime);
 
             GameRulesChangedPacket gameRulesChanged = new GameRulesChangedPacket();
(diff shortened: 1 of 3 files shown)


PleaseInsertNameHere (approved):

src/main/java/cn/nukkit/Player.java
```diff
@@ -4971,8 +4971,8 @@ public boolean switchLevel(Level level) {
                 playerChunkManager.getUsedChunks().clear();
             }
 
-            SetTimePacket setTime = new SetTimePacket();
-            setTime.setTime(level.getTime());
+            final SetTimePacket setTime = new SetTimePacket();
+            setTime.setTime((int) level.getTime());
```
PleaseInsertNameHere, line 4975: Might break at some point

TraxTRL: I have tried it, it breaks on 2.0.0 but not on 3.0.0. Seems like the Cloudburst protocol implementation handles it different. I had a version where it wrapped the int within the 32-bit range, but that's not needed since the client already handles/corrects it.

TraxTRL: Side effect of that wrap was that the client was cycling around day 1-7 on the days played counter, that's why I dropped it (client shows `Too many to count!` now).

PR #2617, opened 2026-06-23
fix!: stop block-entity duplication & make chunk auto-save consistent

TraxTRL:
Continuation of PR #2600 since it wouldn't let me rebase without requiring me to resolve multiple dead conflicts.

---

Fixes a family of block-entity corruption bugs that traced back to a single root cause: chunks could be initialized more than once concurrently, creating duplicate block entities at the same position. This surfaced as:

- a `ClassCastException` that halted an entire level's tick loop (a `BlockEntityCrafter` left orphaned at a position now occupied by a hopper, `(BlockCrafter) getLevelBlock()` on a `BlockHopper`),
- signs (and other block entities) reverting to stale content after a restart,
- generally inflated/duplicated block-entity data accumulating rapidly near players.

It also hardens the chunk auto-save path, which independently could write a torn block/block-entity snapshot.

`Chunk#initChunk` did an unsynchronized check-then-act on `isInit`. `forceLoadChunk` is reachable concurrently from several threads (the tick thread, the async chunk loader, netty login and entity AI (`asyncPrepare` → `getBlock` → `getChunk`)), all operating on the same chunk instance. Multiple threads entering `initChunk` each re-ran the body and created a full duplicate set of the chunk's (block) entities.

Separately, auto-save serialized chunks on a background thread while the main thread kept mutating them, so the block palette and the block-entity set could be captured at different instants.

### Changes

- `Chunk#initChunk` now runs exactly once per chunk via double-checked locking (`isInit` is `volatile`, body guarded by a dedicated lock)
- Chunks are now serialized on each level's tick thread (the sole mutator) a bounded number per tick (see config) and only the LevelDB write is async (single-threaded per-provider writer, drained on close so nothing is lost on shutdown). This removes the torn-snapshot race
- New `saveChunksPerTick` (default `16`) under base settings, next to `autoSave`/`autosaveDelay`. Lower spreads a save over more ticks (smoother TPS); higher finishes sooner. Read live each tick
- The chunk serializer deduplicates block entities by position, preferring the live (position-registered) entity, and warns if a duplicate is ever seen again so pre-existing on-disk corruption is collapsed on save and regressions are visible
- `CrafterInventory` no longer hard-casts `getLevelBlock()` to `BlockCrafter`; it updates redstone only when the block really is a crafter and otherwise closes the orphaned block entity (heals a block↔block-entity type mismatch instead of throwing during tick)

### API changes

- `CrafterInventory#getComperatorOutput` was renamed to `getComparatorOutput` (typo)

Note for reviewers: This is mostly made with Claude Opus 4.8, so please be extra cautious while reviewing. I have tested those changes (even in production), but I still don't 100% trust AI with level related stuff.


TraxTRL:
This is running in production for about 5 days now and so far no issues have been reported.

This still needs testing with levelthreads enabled though.

TraxTRL:
I will revise this PR and split it into multiple parts, closing for now.

PR #2641, opened 2026-06-24
fix: derive player UUID from XUID instead of deprecated identity claim

TraxTRL:
This PR replaces the deprecated identity claim usage with the same UUID deriving method that's already used in the Player creation. Introcoduces a new helper method `Server#uuidFromXUID(String)`

### Compatibility / existing data

`Player.uuid` is unchanged by this PR (XUID-derived before and after), so player data are not re-keyed and nothing is lost.

Servers that previously ran a build where the deprecated `identity` was used as the data key may have leftover name->UUID index entries pointing at `identity`-derived UUIDs. Those UUIDs hold no data (`hasOfflinePlayerData` returns false for them), but I'm not 100% if they get removed automatically already.

### AI usage

Claude has been used to generate Javadoc for the new helper method.


tunikakeks (approved):

xRookieFight (approved):
LGTM

PR #2680, opened 2026-07-01
fix: register client-side multi recipes

TraxTRL:
Allows crafting of shield banners, map cloning, fireworks etc.

If a recipe has no crafting block, it's a multi recipe (type 4). These are resolved client-side, but only once the server registers them and includes them in the crafting packet, previously they were skipped entirely. We now register them and on craft pass the recipe through the request context so the result the client sends is applied.

Claude Opus 4.8 was used to identify the cause and suggest a fix. The suggested fix has been applied with some minor changes.


AzaleeX (approved):

tunikakeks (approved):

TraxTRL:
https://github.com/PowerNukkitX/PowerNukkitX/blob/040d4a280f778f880b2f99cb89a68303af6d5dcf/src/main/java/cn/nukkit/registry/RegistryCache.java#L48

Does this need a bump?

PleaseInsertNameHere:
No

PR #2686, opened 2026-07-02
Fix loading of images into ItemFilledMap.

JBent90914:
This fixes the loading of images into ItemFilledMaps. Previously, they were axis-swapped.
The second change fixes a piece of commit acddcea. There's no obvious reason given for why the retransmit was originally added. It sent the map a second time, after a delay of one second. Clients apparently have never complained until iOS 26.3x.


PleaseInsertNameHere (approved):
Thank you! 😀

tunikakeks (approved):

PR #2688, opened 2026-07-04
Update Dockerfile to Temurin and CI-safe build

AzaleeX:
Switches both build and runtime stages from OpenJDK slim images to Eclipse Temurin, updates build inputs (including gradle.properties and gradlew.bat), and hardens package installation with no-recommends plus apt cache cleanup. The build now runs Gradle in non-daemon, no-configuration-cache mode and copies the generated jar from a stable path. It also sets non-interactive setup mode for container startup and clarifies volume purpose in comments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,28 +2,32 @@
 # See https://docs.docker.com/engine/userguide/eng-image/multistage-build/
 # Requires Docker v17.05
 
-# Use OpenJDK JDK image for intermiediate build
-FROM openjdk:21-slim AS build
+# Use Temurin JDK image for intermediate build
+FROM eclipse-temurin:21-jdk AS build
 
 # Build from source and create artifact
 WORKDIR /src
 
-COPY gradlew *.gradle.kts .gitmodules /src/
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,28 +2,32 @@
...
+    && apt-get install -y --no-install-recommends git \
+    && rm -rf /var/lib/apt/lists/*
 RUN git submodule update --init
-RUN ./gradlew shadowJar
+RUN ./gradlew shadowJar --no-daemon --no-configuration-cache
 
-# Use OpenJDK JRE image for runtime
-FROM openjdk:21-slim AS run
+# Use Temurin JRE image for runtime
+FROM eclipse-temurin:21-jre AS run
```
PleaseInsertNameHere: Using JRE previously caused issues; we should use JDK here

PleaseInsertNameHere (changes requested):

PleaseInsertNameHere (approved):

tunikakeks (approved):

PR #2693, opened 2026-07-06
fix: Fire protection stacking not preventing ignite on players

sircurse:
Fix fire protection stacking not preventing ignite on players.

### DIFF
--- a/src/main/java/cn/nukkit/entity/EntityHumanType.java
+++ b/src/main/java/cn/nukkit/entity/EntityHumanType.java
@@ -136,17 +136,31 @@ public abstract class EntityHumanType extends EntityCreature implements IHuman, 
 
     @Override
     public void setOnFire(int seconds) {
-        int level = 0;
+        if (this.inventory == null) {
+            super.setOnFire(seconds);
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


src/main/java/cn/nukkit/entity/EntityHumanType.java
```diff
@@ -136,17 +136,31 @@ protected double calculateEnchantmentProtectionFactor(Item item, EntityDamageEve
 
     @Override
     public void setOnFire(int seconds) {
-        int level = 0;
+        if (this.inventory == null) {
```
PleaseInsertNameHere: This check is not very plausible, can we use something else?

PleaseInsertNameHere (changes requested):

src/main/java/cn/nukkit/entity/EntityHumanType.java
sircurse: Removed it and kept the logic directly based on the armor contents.

PleaseInsertNameHere (approved):

xRookieFight (approved):

PR #2702, opened 2026-07-08
feat: add FoodEatEvent with interact bypass support

SemihAkin685:


xRookieFight (changes requested):
We can remove the event and bring the bypass thing into "PlayerConsumeEvent"

xRookieFight (approved):

PleaseInsertNameHere (approved):

PR #2712, opened 2026-07-09
Refactoring CustomItem and supporting Legacy Items

Zwuiix-cmd:


Buddelbubi:
Why legacy items? Do they even work on the latest client?

Zwuiix-cmd:
Of course, these have been around since the game's inception, have never changed, and won't change since Minecraft still runs on them; moreover, legacy items have no issues with textures or animations, unlike data-driven ones.

SenseiTarzan:
> Of course, these have been around since the game's inception, have never changed, and won't change since Minecraft still runs on them; moreover, legacy items have no issues with textures or animations, unlike data-driven ones.

Please note that Legacy Mode is a Minecraft feature, and there aren't many items left in it; it may be removed in the near future, and many of the new items are now data-driven.

xRookieFight:
> Please note that Legacy Mode is a Minecraft feature, and there aren't many items left in it; it may be removed in the near future, and many of the new items are now data-driven.

Honestly, I find this comment pretty accurate because Mojang introduced data-driven items in 26.40 so probably they will remove it. The PR status should be decided by @PowerNukkitX/core-dev, maybe?

AzaleeX:
The problem is that legacy items don't have a weird tap animation, unlike data-driven items, which have a broken animation when you tap them.

SenseiTarzan:
> The problem is that legacy items don't have a weird tap animation, unlike data-driven items, which have a broken animation when you tap them. 

Note that this can be fixed with custom animations on the item, and that solves the problem of the strange item animation.

Zwuiix-cmd:
> > Of course, these have been around since the game's inception, have never changed, and won't change since Minecraft still runs on them; moreover, legacy items have no issues with textures or animations, unlike data-driven ones.
> 
> Please note that Legacy Mode is a Minecraft feature, and there aren't many items left in it; it may be removed in the near future, and many of the new items are now data-driven.

Wait, what? What you're saying is totally wrong. Yes, Minecraft uses data-driven items, but only for the items that require it. Otherwise, copper armor and lots of other items would have been made data-driven, which isn't the case. Minecraft has no reason to remove legacy items, and it’s practically impossible for them to do so for the simple reason that every single item in the game is based on them—all basic items are legacy items, and The Hive also uses legacy items. The idea that data-driven will replace legacy items is a common misconception and has no basis in fact.

PR #2725, opened 2026-07-12
fix: Show damage for custom tools

iYozemMc:


src/main/java/org/powernukkitx/item/customitem/CustomItemDefinition.java
```diff
@@ -1086,8 +1086,7 @@ private void writeDamage(CompoundTag itemProps, CompoundTag components) {
             if (this.damage == null) return;
 
             itemProps.putInt("damage", this.damage);
-            components.putCompound("minecraft:damage",
-                    new CompoundTag().putByte("value", this.damage.intValue() & 0xFF));
+            components.putByte("minecraft:damage", this.damage & 0xFF);
```
sircurse: I tested this against a real BDS item definition using `minecraft:damage: { "value": 20 }`.
BDS sends:
```text
"item_properties": {
    "damage": 20i
},
"minecraft:damage": {
    "value": 20s
}
```

So `minecraft:damage` should remain a compound, and the nested `value` is a short. This PR changes it to a direct byte, which does not match the actual packet structure. The fix should likely use `putShort("value", ...)` instead.

tunikakeks: @iYozemMc

sircurse (approved):

PleaseInsertNameHere (approved):

PR #2726, opened 2026-07-12
fix: Make BlockBubbleColumn motion modifier vanilla

iYozemMc:


tunikakeks (approved):

TraxTRL:
Currently looking into it and it's still bugged. I will try to find out why

src/main/java/org/powernukkitx/block/BlockBubbleColumn.java
```diff
@@ -18,9 +18,12 @@
 import javax.annotation.Nullable;
 import java.util.concurrent.ThreadLocalRandom;
 
-
 public class BlockBubbleColumn extends BlockTransparent {
     public static final BlockProperties PROPERTIES = new BlockProperties(BUBBLE_COLUMN, CommonBlockProperties.DRAG_DOWN);
+    private static final double BUBBLE_UP_TARGET_SPEED = 8.0 / 20.0;
```
TraxTRL: Increase from 8.0 to 14.0. Minecraft Wiki seems to be wrong about this.

src/main/java/org/powernukkitx/block/BlockBubbleColumn.java
```diff
@@ -106,32 +109,31 @@ protected AxisAlignedBB recalculateBoundingBox() {
...
-                }
-                
-                ThreadLocalRandom random = ThreadLocalRandom.current();
-                for(int i = 0; i < 2; ++i) {
-                    level.addParticle(new WaterSplashParticle(add(random.nextFloat(), random.nextFloat() + 1, random.nextFloat())));
-                    level.addParticle(new BubbleParticle(add(random.nextFloat(), random.nextFloat() + 1, random.nextFloat())));
-                }
-                
+        if (!entity.canBeMovedByCurrents()) return;
+
```
TraxTRL, line 124: Return if entity instanceof Player and let the client handle it

src/main/java/org/powernukkitx/block/BlockBubbleColumn.java
```diff
@@ -106,32 +109,31 @@ protected AxisAlignedBB recalculateBoundingBox() {
...
+                entity.motionY = Math.min(BUBBLE_UP_TARGET_SPEED, entity.motionY + BUBBLE_ACCELERATION);
             }
-            entity.motionChanged = true;
-            entity.resetFallDistance();
+        }
+
+        if (up().isAir()) {
+            spawnColumnParticles();
+        }
+
```
TraxTRL: Move this to the top after the canBeMovedByCurrents check

src/main/java/org/powernukkitx/block/BlockBubbleColumn.java
```diff
@@ -106,32 +109,31 @@ protected AxisAlignedBB recalculateBoundingBox() {
...
+            if (entity instanceof EntityPhysical entityPhysical && entity.motionY < -entityPhysical.getGravity() * 8) {
+                entity.motionY = -entityPhysical.getGravity() * 2;
             } else {
-                if (isDragDown()) {
-                    entity.motionY = Math.max(-0.3, entity.motionY - 0.3);
-                } else {
-                    entity.motionY = Math.min(0.7, entity.motionY + 0.06);
-                }
+                entity.motionY = Math.min(BUBBLE_UP_TARGET_SPEED, entity.motionY + BUBBLE_ACCELERATION);
             }
```
TraxTRL: Overall, this still seems a bit messy. Entities are being launched out of the water without loosing height or any cap. See Minecraft Wiki for reference.

> At high enough speeds, players and entities can be launched out of water by several blocks. Sitting on the surface above a rising bubble column causes entities to bounce upward repeatedly by about 1 block.
https://minecraft.wiki/w/Bubble_Column#Upward

TraxTRL (changes requested):

src/main/java/org/powernukkitx/block/BlockBubbleColumn.java
```diff
@@ -106,32 +110,28 @@ protected AxisAlignedBB recalculateBoundingBox() {
...
-            } else {
-                if (isDragDown()) {
-                    entity.motionY = Math.max(-0.3, entity.motionY - 0.3);
-                } else {
-                    entity.motionY = Math.min(0.7, entity.motionY + 0.06);
-                }
-            }
-            entity.motionChanged = true;
-            entity.resetFallDistance();
+        if (entity instanceof Player) return;
```
TraxTRL: Move this below the up().isAir() check so players also trigger spawnColumnParticles()

TraxTRL (changes requested):

src/main/java/org/powernukkitx/block/BlockBubbleColumn.java
```diff
@@ -106,32 +110,28 @@ protected AxisAlignedBB recalculateBoundingBox() {
...
-                }
-            }
-            entity.motionChanged = true;
-            entity.resetFallDistance();
+        if (up().isAir()) {
+            spawnColumnParticles();
+        }
+
+        if (entity instanceof Player) return;
+        if (!(entity instanceof EntityItem) && !entity.canBeMovedByCurrents()) return;
```
TraxTRL: spawnColumnParticles should only trigger after the canBeMovedByCurrents check and the EntityItem check is redundant.

```java
        if (!entity.canBeMovedByCurrents()) return;

        if (up().isAir()) {
            spawnColumnParticles();
        }

        if (entity instanceof Player) return;
```

src/main/java/org/powernukkitx/block/BlockBubbleColumn.java
```diff
@@ -18,9 +18,13 @@
 import javax.annotation.Nullable;
 import java.util.concurrent.ThreadLocalRandom;
 
-
 public class BlockBubbleColumn extends BlockTransparent {
     public static final BlockProperties PROPERTIES = new BlockProperties(BUBBLE_COLUMN, CommonBlockProperties.DRAG_DOWN);
+    private static final double DOWNWARD_MIN_MOTION = -0.3;
+    private static final double DOWNWARD_ACCELERATION = 0.03;
+    private static final double UPWARD_ACCELERATION = 0.08;
+    private static final double UPWARD_MAX_MOTION = 0.6;
```
TraxTRL: I might be wrong, but the name of DOWNWARD_MIN_MOTION and UPWARD_MAX_MOTION is wrong, isn't it? After looking at its usage, I think it should be DOWNWARD_MAX_MOTION and UPWARD_MIN_MOTION.

TraxTRL (dismissed):

tunikakeks (approved):

TraxTRL:
This is an important fix (e.g. item transport systems and elevators depend on this) so please don't remove the high priority label again. Thank you!

TraxTRL:
Traced the packets on both PNX and BDS to try and match the behavior just to get the boat shake working, but for some reason it is still not working. As it is just a visual, I don't want make this halt our PR. This is good as it is now.

PleaseInsertNameHere (approved):

PR #2754, opened 2026-07-16
Harden packet handlers, drop unused handlers

Zwuiix-cmd:
fixed somes exploits and added todo for unknown patch


src/main/java/org/powernukkitx/network/process/PacketHandlerRegistry.java
```diff
@@ -43,11 +43,11 @@ public class PacketHandlerRegistry {
         register(MobEquipmentPacket.class, new MobEquipmentHandler());
         register(ModalFormResponsePacket.class, new ModalFormResponseHandler());
         register(MoveActorAbsolutePacket.class, new MoveActorAbsoluteHandler());
-        register(MovePlayerPacket.class, new MovePlayerHandler());
+        //register(MovePlayerPacket.class, new MovePlayerHandler()); Unused since idk
         register(NpcRequestPacket.class, new NpcRequestHandler());
-        register(PacketViolationWarningPacket.class, new PacketViolationWarningHandler());
+        //register(PacketViolationWarningPacket.class, new PacketViolationWarningHandler()); Unused?
```
tunikakeks, line 47: We should not remove PacketViolationWarningPacket processer

Zwuiix-cmd: The issue with this packet is that an attacker can spam it, causing the server to freeze because of the spam generated in the console.

Psycofeu: Exploit creator btw

Buddelbubi: Im with keks here. Instead of removing this, we could disconnect the player if sent too often.
This packet usually is not sent

Zwuiix-cmd: I agree, but if we take [InvMenu](https://github.com/Muqsit/InvMenu) as an example since it uses it as an ACK we'd need to implement a proper solution.

tunikakeks: We could handle the packet with debug mode enabled as it's usually not needed anyways, what do you think @Buddelbubi

xRookieFight:
Thank you for taking the time to contribute.

Unfortunately, we're going to close this pull request. Throughout the review process, we decided that this PR is not really fixing the issue, making it even worse.

We appreciate your interest in contributing. If you'd like to submit another pull request in the future, please make sure to carefully test the PR, follow the project's conventions, and ensure that the PR description is fully understood, reviewed, and adapted before submission.

PR #2755, opened 2026-07-17
fix: Rabbit not being vanilla

Buddelbubi:


PleaseInsertNameHere (approved):

src/main/java/org/powernukkitx/entity/ai/controller/HoppingController.java
```diff
@@ -9,6 +10,8 @@
 import org.cloudburstmc.protocol.bedrock.data.actor.ActorDataTypes;
 import org.cloudburstmc.protocol.bedrock.data.actor.ActorFlags;
 
+import java.util.Arrays;
+
 /**
  * 处理陆地行走实体运动
  * todo: 有待解耦
```
tunikakeks: please translate those comments too

tunikakeks (approved):

PR #2759, opened 2026-07-17
Optimize villager AI performance via POI system

SenseiTarzan:
## Description
Implemented an initial version of a POI (Point of Interest) system for villagers, aimed at improving AI performance.

## Problem
Currently, once more than 15 villagers are present at the same time, server TPS drops significantly due to the computational cost of villager AI (pathfinding, interactions, etc.).

## Solution
Introduced a POI system that:
- Reduces the computational load related to villager behaviors
- Prevents TPS drops even with a high number of villagers
- Lays a scalable foundation for future AI optimizations

## Current status
This is an initial implementation (WIP). Only a base version of the POI system has been added for now.

## Future work
- Implement additional POI types to further refine and expand villager AI behaviors
- Continue optimizing performance based on testing results

## Testing done
https://youtu.be/DMt2bIcNBlA?is=jLDvw7AeWN8Wgen0


Buddelbubi:
Are POI's vanilla?

SenseiTarzan:
Yes, POIs are a technology used in vanilla Minecraft too; they serve many purposes, and I've added the base ones.

SenseiTarzan:
It drastically narrows down the search, because only blocks associated with professions are included in that section; this means we have fewer blocks to search through—such as the bed, other blocks, and so on.

Buddelbubi:
Why do you need a level wide PoiManager? For me it looks like its chunk based. Like it splits it into chunks internally. Also when saving. Is there a reason I dont see why it has to be in the level?

SenseiTarzan:
In this case, I kept them separate because I followed the basic concept of vanilla Minecraft.

Buddelbubi:
Is that scanning vanilla? Bacause scanning a chunk is not good for performance. We had major issues with that before.
Also what happens when you break a POI, like a bed?

Buddelbubi:
> Is that scanning vanilla? Bacause scanning a chunk is not good for performance. We had major issues with that before. Also what happens when you break a POI, like a bed?

Oh I see.

src/main/java/org/powernukkitx/level/format/leveldb/LevelDBChunkSerializer.java
```diff
@@ -74,6 +74,10 @@ public void serialize(WriteBatch writeBatch, IChunk chunk) {
             serializeHeightAndBiome(writeBatch, unsafeChunk);
             serializeLight(writeBatch, unsafeChunk);
             serializeBlockTicks(unsafeChunk);
+            if (unsafeChunk.getProvider() != null && unsafeChunk.getProvider().getLevel() != null) {
```
Buddelbubi: Please do that in a seperate method

src/main/java/org/powernukkitx/level/format/palette/Palette.java
```diff
@@ -57,6 +57,15 @@ public Palette(V first, List<V> palette, BitArrayVersion version) {
         this.addToPalette(first);
     }
 
+    public boolean anyMatch(java.util.function.Predicate<V> predicate) {
+        for (int i = 0, size = this.palette.size(); i < size; i++) {
+            if (predicate.test(this.palette.get(i))) {
+                return true;
```
Buddelbubi, line 64: public boolean anyMatch(Predicate<V> predicate) {
        for (V v : this.palette) {
            if (predicate.test(v)) {

Buddelbubi (changes requested):
Also if its vanilla, isn't there an actual key for that? Because right now, its written to extra data

src/main/java/org/powernukkitx/level/format/palette/Palette.java
xRookieFight: Also please import the Predicate

SenseiTarzan: Okay, I'll do that.

src/main/java/org/powernukkitx/level/format/leveldb/LevelDBChunkSerializer.java
SenseiTarzan: ~~I just retrieved the key levelDB to set the POI~~
I just watched the official explanation; it's very poorly explained. I'll have to reverse engineer an official Minecraft map to find the prefix or key to store all of this.

SenseiTarzan: I just checked, it's Java that saves the POI, and in Bedrock Edition it constantly regenerates with each chunk load.

SenseiTarzan:
I can offer you the option of either leaving the POI saved or doing it like Bedrock; it's up to you.

Buddelbubi:
> I can offer you the option of either leaving the POI saved or doing it like Bedrock; it's up to you.

If you could do it like bedrock, that would be awesome. As I know its saved to the DB with a global key. Like VILLAGE_OVERWORLD_{village_id},
But we have no system for that rn.
But then we could also store village bounding boxes, which would allow us to implement raids

SenseiTarzan:
And I understand that you want me to set up a global key containing all the information from the POIManager.

SenseiTarzan:
We can create an initial version that's dynamic and rescans the chunk every time it's loaded, and then I'll submit another pull request for the backup feature, which will include part of the optimization.

Buddelbubi:
> We can create an initial version that's dynamic and rescans the chunk every time it's loaded, and then I'll submit another pull request for the backup feature, which will include part of the optimization.

Yea. That would be better for now. Saving vanilla stuff to chunk extras is not good in my opinion

Buddelbubi:
@SenseiTarzan Any news on this?

SenseiTarzan:
> @SenseiTarzan Any news on this?

I get to my PC and I do this today, I delete the save from the POI and I rescan every time the chunk loads

Buddelbubi:
> > @SenseiTarzan Any news on this?
> 
> I get to my PC and I do this today, I delete the save from the POI and I rescan every time the chunk loads

okay. Please also check if it has an noticable negative impact on performance. If it does, I will implement the correct way to store them first

Buddelbubi:
> > @SenseiTarzan Any news on this?
> 
> I get to my PC and I do this today, I delete the save from the POI and I rescan every time the chunk loads

I just checked how villages are actually saved. Is probably cleaner to do an VillageManager instead of just doing POIs. Writing the other village stuff afterwards would be pretty difficult

SenseiTarzan:
After testing, I drop to 5 TPS (constant) with the old code; with the new code, I drop to 15 TPS but bounce back to a steady 20 TPS after the scan.
Old system search:
Screenshot From 2026-07-21 15-27-30
New system search: 
image

Buddelbubi:
Wait a second. I will send something shortly.

Buddelbubi:
https://github.com/PowerNukkitX/PowerNukkitX/commit/d9b29a3f0b561d20bd25232ab356fe9468a2258b

@SenseiTarzan would you check this?
This should be even better

SenseiTarzan:
> [d9b29a3](https://github.com/PowerNukkitX/PowerNukkitX/commit/d9b29a3f0b561d20bd25232ab356fe9468a2258b)
> 
> @SenseiTarzan would you check this? This should be even better

I think it would be more sensible to keep the `VillageUuid` reference directly in `EntityVillage` to avoid going through a `HashMap`, and I’m talking about the `CompoundTag` here; this would reduce latency by avoiding the key’s hash computation.
https://github.com/PowerNukkitX/PowerNukkitX/blob/d9b29a3f0b561d20bd25232ab356fe9468a2258b/src/main/java/org/powernukkitx/entity/passive/EntityVillagerV2.java#L788-L801

You can also tell him that, for raids, it would still be more interesting to keep even the Village, since it would make it possible to check whether a raid is currently happening there.

And don’t forget that POIs can exist outside of a village; that’s why Minecraft Java keeps the POI system separate, since you can create special villagers on blocks without them being tied to a village.

SenseiTarzan:
Otherwise, the village system you implemented is really cool and great, and we also need to take into account the fact that we can create villages by placing linked blocks and doors. I don’t remember whether it works the same way on Minecraft Bedrock, and that allows us to make iron farms.

Buddelbubi:
> Otherwise, the village system you implemented is really cool and great, and we also need to take into account the fact that we can create villages by placing linked blocks and doors. I don’t remember whether it works the same way on Minecraft Bedrock, and that allows us to make iron farms.

If you want, you can PR into that branch and fix those stuff. Ik that village detection is not vanilla rn. Its documented in Minecraft Wiki how its done. That would also allow to create villages by yourself

SenseiTarzan:
> If you want, you can PR into that branch and fix those stuff. Ik that village detection is not vanilla rn. Its documented in Minecraft Wiki how its done. That would also allow to create villages by yourself

I’m going to add the artificial villages and open a new PR on the `feat/villageData` branch.

Buddelbubi:
> > If you want, you can PR into that branch and fix those stuff. Ik that village detection is not vanilla rn. Its documented in Minecraft Wiki how its done. That would also allow to create villages by yourself
> 
> I’m going to add the artificial villages and open a new PR on the `feat/villageData` branch.

SUre. Btw other POIs dont exist in save data, Therefore I'd assume that other POIs are only in Java Edition

Buddelbubi:
https://minecraft.wiki/w/Village#Bedrock_Edition_villages

Buddelbubi:
https://github.com/PowerNukkitX/PowerNukkitX/pull/2826

PR #2806, opened 2026-07-21
fix: preserve unknown block identity in chunk palette and improve logging

TraxTRL:
This PR fixes the block state updater so Unknown blocks no longer loose the "Block" identity when saved to disk.
Additionally, it improves logging by adding position, changing the wording a bit, and dropping empty blocks without recoverable identity to prevent log spam.
Finally, it also adds a test to verify that the block state updater is able to recover blocks.

To test this in-game, install a plugin implementing a custom block, place it on a world, restart the server without the plugin, force the chunk to get saved to disk, restart the server with the plugin again. If the block recovers, the changes work. `saveUnknownBlock` must be enabled in `pnx.yml` for this to work. Also set your log level to DEBUG if you want to verify them too.

AI usage: [Test](https://github.com/PowerNukkitX/PowerNukkitX/commit/7e48092c4ec3094c1eef83d0efd27775b66e690b) by Claude Opus 4.8

### DIFF
--- a/src/main/java/org/powernukkitx/level/format/leveldb/LevelDBChunkSerializer.java
+++ b/src/main/java/org/powernukkitx/level/format/leveldb/LevelDBChunkSerializer.java
@@ -47,6 +47,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
+import java.util.function.Supplier;
 
 /**
  * Allay Project 8/23/2023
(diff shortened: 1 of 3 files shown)


TraxTRL:
I tested it in-game and this is working; Unknown blocks now recover! 🎉

xRookieFight (approved):
LGTM

tunikakeks (approved):

PR #2818, opened 2026-07-21
Fix memory leaks, stability bugs and hot path performance

calci123:
Bugs:
- EntityPhysical.handleCollideMovement: unsynchronized writes to a non-thread-safe DoubleArrayList from a parallel stream (data race). Replaced with a sequential loop tracking four running maxima instead of four lists. Added a NaN guard for the len == 0 case where symmetric pushers cancel out.
- BlockEntityChest / BlockEntitySpawnableContainer.close: ConcurrentModificationException while iterating viewers aborted close() and left windows open. Use a defensive copy, matching the existing pattern in ShulkerBox and Hopper.
- Level.syncGenerateChunk: a bare spin loop and a leaked generation queue entry could permanently stall all chunk generation. Bounded 30s wait via ForkJoinPool.ManagedBlocker, try/finally entry release, and a null guard on the recovery path. Generator.asyncGenerate0 now always invokes the callback, including on stage failure and rejected execution.

Leaks:
- Level.unregisterChunkLoader: an early return skipped the loaders/loaderCounter cleanup, retaining Player instances for the lifetime of the server.
- Network.initSession: sessionMap entries were never removed on disconnect. Added a closeFuture listener with conditional removal so a stale session cannot evict a newer one bound to the same address.
- Level.doLevelGarbageCollection: the metadata key was disambiguated twice, so the lookup never matched and block metadata was never collected.
- Level.remove now closes the per-level scheduler; ServerScheduler.close performs a graceful shutdown so in-flight tasks drain before the provider closes.
- Level.releaseTickCachedBlocks: drop tick cache entries that went unused.

Hot paths:
- Nearest{Entity,Item,TargetEntity}Sensor: scan a chunk window instead of copying and walking the level-wide entity array on every tick.
- Level.getChunkAsync: return a completed future directly when the chunk is already resident instead of dispatching to the async pool.
- Level.updateBlockSkyLight: resolve the chunk once and read block states directly, removing 384 Block allocations per call, and break once the light level clamps.
- BlockProperties.containBlockState: replace the O(n) containsValue scan with an O(1) key lookup.
- Level.tickChunk: run the entity update loop synchronously instead of submitting a task to the common pool every tick.
- PlayerChunkManager: hoist loop-invariant position and FOV computations out of the chunk queue comparator.
- AsyncPool: raise keepAlive from 60ms to 30s so workers survive between ticks, and enable allowCoreThreadTimeOut.

## AI usage

This PR was produced with AI assistance for the analysis and the initial patches.
Every change was reviewed before submitting, and the build was verified and
runtime-tested on a local server.

### DIFF
--- a/src/main/java/org/powernukkitx/block/BlockProperties.java
+++ b/src/main/java/org/powernukkitx/block/BlockProperties.java
@@ -142,7 +142,9 @@ public final class BlockProperties {
     }
 
     public boolean containBlockState(BlockState blockState) {
-        return this.specialValueMap.containsValue(blockState);
+        if (blockState == null) return false;
+        BlockState canonical = this.specialValueMap.get(blockState.specialValue());
+        return canonical != null && canonical.equals(blockState);
     }
 
     public boolean containBlockState(short specialValue) {
--- a/src/main/java/org/powernukkitx/blockentity/BlockEntityChest.java
+++ b/src/main/java/org/powernukkitx/blockentity/BlockEntityChest.java
@@ -1,5 +1,6 @@
 package org.powernukkitx.blockentity;
 
+import org.powernukkitx.Player;
 import org.powernukkitx.block.BlockChest;
 import org.powernukkitx.block.copper.chest.BlockCopperChest;
 import org.powernukkitx.inventory.BaseInventory;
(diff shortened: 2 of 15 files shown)


PleaseInsertNameHere (approved):

xRookieFight (approved):
LGTM

Buddelbubi (approved):

PR #2839, opened 2026-07-23
fix: scatter herd animals across separate positions on spawn

SemihAkin685:


Buddelbubi:
They spawn on the same positition. At least in java. I could not validate it in bedrock. I added a minimal shift so they drift apart and wont push each other

xRookieFight (approved):

tunikakeks (approved):

PR #2849, opened 2026-07-23
fix: player invisible after respawn when killed with UI open

SemihAkin685:


src/main/java/org/powernukkitx/Player.java
```diff
@@ -2900,11 +2900,18 @@ public boolean onUpdate(int currentTick) {
 
         if (!this.isAlive() && this.spawned) {
             this.drainInboundPackets();
+            if (this.isAlive()) {
```
Buddelbubi, line 2903: What? There is a negative check before

SemihAkin685: The check is not redundant: `drainInboundPackets()` on the line above can process the client's `RESPAWN` player action, which calls `respawn()` and brings the player back to life mid-tick. Without re-checking, the branch would still schedule a `despawnFromAll()` task 10 ticks later, sending `RemoveActorPacket` to all viewers *after* `spawnToAll()` already ran — which is exactly the invisible-after-respawn bug this PR fixes. Same reason the delayed task body re-checks `isAlive()`: tasks scheduled in earlier dead ticks can fire after the respawn.

Buddelbubi (approved):

Buddelbubi:
@SemihAkin685 Please merge master into this

xRookieFight (approved):
LGTM

PR #2850, opened 2026-07-23
fix: NPE on block entity setting dirty

xRookieFight:


tunikakeks (approved):

src/main/java/org/powernukkitx/blockentity/BlockEntity.java
```diff
@@ -214,7 +214,13 @@ public void onBreak(boolean isSilkTouch) {
     }
 
     public void setDirty() {
-        chunk.setChanged();
+        if (chunk != null) {
+            chunk.setChanged();
+        }
+
+        if (this.level == null) {
```
Buddelbubi: replace with isValid()

Buddelbubi (dismissed):

PleaseInsertNameHere (approved):

PR #2851, opened 2026-07-23
perf!: optimize hot paths in level, entity, village and inventory systems

SemihAkin685:
This PR reduces per-tick allocations, removes stream pipelines from hot paths, and eliminates redundant work across the level, entity, village and inventory systems. No behavior changes intended; all existing tests pass.

## Level / chunk ticking
- `Level#getChunkPlayers` no longer builds a stream + collector per call; added `Level#getChunkPlayerArray` which fills a plain array directly. The `doTick()` changed-blocks and chunk-packet loops and the direct `setBlock` path (which previously called `getChunkPlayers` three times for the same chunk) now use it.
- `updateBlockLight` no longer clones the pending `IntOpenHashSet` per chunk — the snapshot map is exclusively owned after the synchronized copy, so the sets are iterated directly.
- `chunkPackets` broadcast loop iterates `entrySet()` instead of `keySet()` + `get()`.
- `getNearbyEntitiesSafe` and `setSpawnLocation` use plain loops instead of stream pipelines.

## Entity ticking
- `Entity#move` reuses two preallocated AABBs instead of calling `boundingBox.clone()` twice per movement.
- `updateMovement` compares motion components directly instead of allocating a `Vector3` via `getMotion().lengthSquared()`.
- Auto-despawn scan exits early once a player within soft-despawn range is found, instead of always scanning all players.
- `checkBlockStepOn` tracks positions in a `LongOpenHashSet` keyed by block hash instead of a `HashSet<Vector3>` (no `Vector3` allocation per block per tick).
- Scaffolding check queries `level.getBlock(x, y, z, false)` directly instead of allocating a `Location` per column.
- `getClientSyncProperties` result is cached and invalidated on property mutation; `sendData` builds the sync data once instead of twice per packet.
- `hasTag`, `createAddEntityPacket` and `syncAttributes` use plain loops instead of streams.
- `despawnFromAll` / chunk-migration despawn no longer copy `hasSpawned` into an `ArrayList` (it is a `ConcurrentHashMap`, safe to iterate while removing).
- `EntityLiving#hasLineOfSight` corridor math uses primitive doubles instead of ~10 intermediate `Vector3` allocations per call.

## Village system
- `getPoiType` no longer streams over all professions on every block change — job-site block IDs are cached in a set, rebuilt only when the profession registry changes.
- `onBlockChange` returns early when neither the previous nor the current block is a POI.
- `isDweller`, `getVillageForDweller`, `getVillageAt`, `findAt`, `removePoi`, `dissolveVillage`, `discoverDwellers` and `recalculateBounds` converted from nested stream pipelines to loops with early exit; `recalculateBounds` accumulates bounds in primitive ints instead of allocating a `BlockVector3` per POI.
- `Profession#getProfessions` returned a fresh `HashMap.clone()` on every call — now returns an unmodifiable view backed by an `Int2ObjectOpenHashMap`.

## Chunk sending
- `PlayerChunkManager` replaces both Guava `Sets.difference` usages (which box every `long` and one of which also made a defensive `HashSet` copy per tick) with primitive `LongIterator` loops.

## Inventory
- `BaseInventory#slots` switched from `HashMap<Integer, Item>` to `Int2ObjectOpenHashMap<Item>` — slot access no longer boxes indices.
- `EntityMob#getDrops` filters with a loop instead of a stream.

## Storage / serialization
- `LevelDBChunkSerializer` dropped the redundant `getNbt().copy()` before `toNetwork()` for both block entities and entities during chunk save (`toNetwork()` is read-only, the deep copy was pure waste).
- `LevelDBStorage` village read/write compares key prefixes as raw bytes and only UTF-8-decodes matching keys.

## Misc
- A* `hasBarrier` (both route finders) and the vibration wool-occlusion check use early-exit loops instead of streams.
- `CameraShakeCommand` sends via `Server.broadcastPacket` instead of a per-player loop.

---

*These changes were made with the help of AI.*


src/main/java/org/powernukkitx/entity/Entity.java
```diff
@@ -241,6 +243,7 @@ public abstract class Entity extends Location implements Metadatable, EntityID {
     protected volatile boolean saveWithChunk = true;
     private final Map<String, Integer> intProperties = new LinkedHashMap<>();
     private final Map<String, Float> floatProperties = new LinkedHashMap<>();
+    private transient PropertySyncData clientSyncPropertiesCache;
```
PleaseInsertNameHere: Callers don't all run on one thread, cross-thread data race

src/main/java/org/powernukkitx/level/village/VillageManager.java
```diff
@@ -32,6 +32,23 @@ public final class VillageManager {
 
     private final Level level;
     private final ConcurrentHashMap<UUID, Village> villages = new ConcurrentHashMap<>();
+    private static volatile java.util.Set<String> jobSiteBlockIds;
+    private static volatile int jobSiteProfessionCount = -1;
```
PleaseInsertNameHere: updated non-atomically
key on a mod-counter bumped in registerProfession instead

src/main/java/org/powernukkitx/level/village/VillageManager.java
```diff
@@ -32,6 +32,23 @@ public final class VillageManager {
...
+
+    private static java.util.Set<String> getJobSiteBlockIds() {
+        var professions = Profession.getProfessions();
+        var ids = jobSiteBlockIds;
+        if (ids == null || jobSiteProfessionCount != professions.size()) {
```
PleaseInsertNameHere: Re-registering an existing index leaves the size unchanged, so the cached block-ID set never rebuilds

src/main/java/org/powernukkitx/level/format/leveldb/LevelDBChunkSerializer.java
PleaseInsertNameHere, line 1: Possible CME, might worth re-checking

PleaseInsertNameHere (changes requested):

src/main/java/org/powernukkitx/level/village/VillageManager.java
```diff
@@ -32,21 +32,23 @@ public final class VillageManager {
...
+        int version = Profession.getRegistrationVersion();
+        var cache = jobSiteCache;
+        if (cache != null && cache.version() == version) {
+            return cache.ids();
+        }
+        var rebuilt = new HashSet<String>();
+        for (Profession profession : Profession.getProfessions().values()) {
+            rebuilt.add(profession.getBlockID());
         }
+        var ids = java.util.Set.copyOf(rebuilt);
```
PleaseInsertNameHere: Use imports

src/main/java/org/powernukkitx/level/village/VillageManager.java
```diff
@@ -32,21 +32,23 @@ public final class VillageManager {
 
     private final Level level;
     private final ConcurrentHashMap<UUID, Village> villages = new ConcurrentHashMap<>();
-    private static volatile java.util.Set<String> jobSiteBlockIds;
-    private static volatile int jobSiteProfessionCount = -1;
+    private record JobSiteCache(java.util.Set<String> ids, int version) {
```
PleaseInsertNameHere: Use imports

PleaseInsertNameHere (changes requested):

PleaseInsertNameHere (approved):

tunikakeks (approved):

PR #2855, opened 2026-07-24
fix: sync saturation attribute to client to prevent hunger bar desync

SemihAkin685:


src/main/java/org/powernukkitx/PlayerFood.java
```diff
@@ -184,6 +186,19 @@ public void sendFood(int food) {
         }
     }
 
+    /**
+     * Sends the current saturation level to the player.
+     */
+    public void sendSaturation() {
```
Buddelbubi, line 193: Saturation should be sent in sendFood. Could you please merge those methods?

Buddelbubi (changes requested):

src/main/java/org/powernukkitx/PlayerFood.java
Buddelbubi: Wait. No. Could you just add a send() method that sends both?

Buddelbubi (changes requested):

Buddelbubi (approved):

Buddelbubi:
Could you merge master into this?
Build is failing bc of this

xRookieFight (approved):
LGTM

PR #2858, opened 2026-07-24
fix: restore original items when plugin is disabled

iYozemMc:
## Breaking API change

The class `UnknownItem` has been replaced by `ItemUnknown`.

- **Reason**: Aligns with names of other classes that extend `Item`


src/main/java/org/powernukkitx/item/ItemUnknown.java
```diff
@@ -0,0 +1,22 @@
...
+        super(originalId, meta, count);
+        this.originalId = originalId;
+        if (tags != null && tags.length > 0) {
+            this.setNbtBytes(tags);
+        }
+    }
+
+    public String getOriginalId() {
+        return originalId;
+    }
```
TraxTRL: We already have `getId()` which does the same, drop this

src/main/java/org/powernukkitx/item/ItemUnknown.java
```diff
@@ -0,0 +1,22 @@
...
+
+    public int getMaxStackSize() {
+        return 64;
+    }
```
TraxTRL, line 15: `@Override` is missing

TraxTRL (changes requested):
We now have two unknown item mechanisms living side by side. ItemHelper.read/write still drives the old one, it branches on `item.isNull()` and the `BlockID.UNKNOWN` marker, all of which assumed `Item.get()` returns AIR for an unregistered id. Now that `Item.get()` returns a non-null `ItemUnknown`, those branches are effectively dead. Rather than adding the new path next to the old, let's consolidate `ItemHelper` onto `ItemUnknown` and deprecate/remove `UnknownItem` so we only have one path handling unknown items.

src/main/java/org/powernukkitx/item/ItemUnknown.java
iYozemMc: done

src/main/java/org/powernukkitx/item/ItemUnknown.java
iYozemMc: done

src/main/java/org/powernukkitx/utils/ItemHelper.java
```diff
@@ -19,6 +19,7 @@
 /**
  * @author Kaooot
  */
+
```
TraxTRL: Revert
```suggestion
```

src/main/java/org/powernukkitx/item/Item.java
```diff
@@ -215,7 +215,10 @@ public static Item get(String id, int meta, int count, byte[] tags, boolean auto
         if (item == null) {
             BlockState itemBlockState = getItemBlockState(id, meta);
             if (itemBlockState == null || itemBlockState == BlockAir.STATE) {
-                return Item.AIR;
+                if (id.equals(Block.AIR)) {
+                    return Item.AIR;
+                }
+                return new ItemUnknown(id, meta, count, tags);
```
TraxTRL: Revert; this effectively made `Item.isNull()` always return false.
```suggestion
                return Item.AIR;
```

src/main/java/org/powernukkitx/utils/ItemHelper.java
```diff
@@ -65,36 +65,45 @@ public Item read(CompoundTag tag) {
...
             if (blockState != null) {
-                if (isUnknownBlock) {//restore unknown block item
-                    item = blockState.toItem();
+                if (isUnknownBlock || wasUnknownItem) {
+                    Item resolvedItem = wasUnknownItem
+                        ? Item.get(item.getId(), damage, amount)
+                        : blockState.toItem();
+
+                    item = !(resolvedItem instanceof ItemUnknown) ? resolvedItem : blockState.toItem();
```
TraxTRL:
```suggestion
                    item = resolvedItem != Item.AIR ? resolvedItem : blockState.toItem();
```

src/main/java/org/powernukkitx/utils/ItemHelper.java
```diff
@@ -65,36 +65,45 @@ public Item read(CompoundTag tag) {
         int damage = !tag.contains("Damage") ? 0 : tag.getShort("Damage");
         int amount = tag.getByte("Count");
         Item item = Item.get(name, damage, amount);
+
```
TraxTRL:
```suggestion
        if (item == Item.AIR) {
            item = new ItemUnknown(name, damage, amount, null);
        }
```

TraxTRL (changes requested):

src/main/java/org/powernukkitx/utils/ItemHelper.java
```diff
@@ -65,36 +64,47 @@ public Item read(CompoundTag tag) {
...
         Item item = Item.get(name, damage, amount);
+        if (item == Item.AIR) {
+            item = new ItemUnknown(name, damage, amount, null);
+        }
         Tag tagTag = tag.get("tag");
-        if (!item.isNull() && tagTag instanceof CompoundTag compoundTag && !compoundTag.isEmpty()) {
-            item.setNbt(compoundTag);
+        CompoundTag savedNbt = tagTag instanceof CompoundTag compoundTag && !compoundTag.isEmpty() ? compoundTag : null;
+        if (savedNbt != null) {
+            item.setNbt(savedNbt);
```
TraxTRL: You can inline this
```suggestion
        if (tagTag instanceof CompoundTag compoundTag && !compoundTag.isEmpty()) {
            item.setNbt(compoundTag);
```

TraxTRL (changes requested):

TraxTRL (approved):

PleaseInsertNameHere (approved):

PR #2859, opened 2026-07-24
fix: changed metrics text

Buddelbubi:


src/main/resources/language/eng/lang.json
```diff
@@ -129,7 +129,7 @@
   "pnx.settings.misc.shutdownmessage": "Message sent when the server shuts down",
   "pnx.settings.misc.installspark": "Whether to install the Spark plugin on the server as profiler",
   "pnx.settings.misc.bypassapicheck": "Ignore plugin API check",
-  "pnx.settings.misc.enablemetrics": "Send statistics to developers",
+  "pnx.settings.misc.enablemetrics": "Send statistics to bStats. We kindly ask that you do not disable this feature",
```
xRookieFight: We should not contain a word like "do not disable this feature". If we put this as a setting, this word is non-sense.

xRookieFight (changes requested):

src/main/resources/language/eng/lang.json
Buddelbubi: This is intentional. 
It does not say "Do not disable".
We're just asking the user to keep it enabled

xRookieFight (approved):

TraxTRL (approved):

Issue #2848, opened 2026-07-23
[Bug]: Skeletons do not attack villagers


EvanTechDev:
### Describe the bug

Skeletons do not target or attack villagers.

### To Reproduce

1. Spawn a skeleton.
2. Spawn a villager near the skeleton.
3. Wait and observe the skeleton's behavior.

The skeleton will not attack the villager.

### Expected behavior

Skeletons should detect nearby villagers and attack them according to vanilla Minecraft behavior.

### Screenshots

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/57df65ec-497c-4e51-8038-0adc49b6d0e7" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/d33a9994-393c-4b43-955e-614517640566" />

Buddelbubi:
https://minecraft.wiki/w/Skeleton#Behavior

Skeletons do NOT attack villagers

Buddelbubi:
<img width="2297" height="51" alt="Image" src="https://github.com/user-attachments/assets/3610f037-cd65-4147-a711-1057cf72ebbf" />

EvanTechDev:
my brain was dumb

PR #2862, opened 2026-07-24
fix: skeleton attack villagers

iYozemMc:
#2848


Buddelbubi:
https://minecraft.wiki/w/Skeleton#Behavior

Skeletons do NOT attack villagers

PR #2873, opened 2026-07-25
Fix intermittent chunk load failures and cross-chunk initialization deadlock

skh6075:
## Summary
Fixes two related chunk-loading bugs in `Level.java` that caused chunks to sometimes fail to load, and in rarer cases caused the server to deadlock entirely.

## Bug 1: Silent chunk load failure in `forceLoadChunk`
When the level provider returned a chunk whose `getProvider()` was `null` (e.g. due to corrupted region data or a race during (re)generation/unload), the old code silently called `unloadChunk()` on it and then returned that same broken, un-initiated chunk as if the load had succeeded. Callers had no way to know the load actually failed, so the chunk would appear to exist while never becoming usable — this is what caused chunks to intermittently "not load."

**Fix:** retry the load once when the provider comes back null; if it still fails, throw (when generating) or return `null` (when not) with a clear log message, instead of silently handing back a dead chunk.

## Bug 2: Deadlock between two chunks initializing each other
`Chunk.initChunk()` constructs block entities (e.g. `BlockEntityLectern`) whose constructors perform a redstone update check on neighboring blocks via `Level.getBlock() -> Level.getChunk()`. If that neighboring chunk was also un-initiated, `getChunk()` would recursively call `initChunk()` on it while still holding the current chunk's monitor.

When two neighboring chunks were being initialized concurrently on two different threads, each thread ended up holding its own chunk's lock while waiting on the other's — a classic AB-BA deadlock, confirmed via `jstack` (`Found one Java-level deadlock`).

**Fix:** added a thread-local reentrancy guard (`INITIALIZING_CHUNK_ON_THREAD`). If a thread is already inside a chunk's `initChunk()` and a nested call needs to initialize a different chunk, it now skips the synchronous `initChunk()` call and returns the chunk as-is (block/palette data is already loaded even though block entities/tick lists aren't built yet). The chunk gets properly initialized on the next non-reentrant call, avoiding the deadlock entirely.

## Testing
- Reproduced original chunk load failures and the deadlock via `jstack` thread dumps in a live server.
- Verified with the patch applied that chunks load reliably and no deadlock occurs under the same load pattern.

## Notes
The root cause of Bug 2 ultimately lives in how block entity construction (`BlockEntityLectern` / `RedstoneComponent`) reaches back into neighboring chunks during `Chunk.initChunk()`. This patch prevents the deadlock at the `Level` layer; a deeper fix (e.g. deferring cross-chunk redstone checks out of the chunk-init critical section) could be considered as follow-up work in `Chunk.java` / `BlockEntityLectern.java`.

### DIFF
--- a/src/main/java/org/powernukkitx/level/Level.java
+++ b/src/main/java/org/powernukkitx/level/Level.java
@@ -61,7 +61,6 @@ import org.powernukkitx.level.generator.biome.BiomePicker;
 import org.powernukkitx.level.generator.holder.ObjectHolder;
 import org.powernukkitx.level.particle.DestroyBlockParticle;
 import org.powernukkitx.level.particle.Particle;
-import org.powernukkitx.level.tickingarea.TickingArea;
 import org.powernukkitx.level.util.EntityQueryUtils;
 import org.powernukkitx.level.util.SimpleTickCachedBlockStore;
 import org.powernukkitx.level.util.TickCachedBlockStore;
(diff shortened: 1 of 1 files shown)


Buddelbubi:
You cannot just skip the initialization of chunks...

Please dont PR AI code if you do not understand the impact it can have.

Also please for your next commits, remove unrelated changes like the removal of whitespaces. It makes it harder to review.

skh6075:
I understand about chunk skip. However, the server died without even a log due to a deadlock issue, which is a headache. I hope it will be fixed as soon as possible.

PR #2876, opened 2026-07-26
fix: drowned throw arrows while target not being visible

Buddelbubi:


src/main/java/org/powernukkitx/entity/ai/evaluator/SightEvaluator.java
```diff
@@ -0,0 +1,25 @@
...
+        this.type = type;
+    }
+
+    @Override
+    public boolean evaluate(EntityIntelligent entity) {
+        if (entity.getMemoryStorage().isEmpty(type)) {
+            return false;
+        } else {
+            return entity.hasLineOfSight(entity.getMemoryStorage().get(type));
+        }
```
PleaseInsertNameHere: Do not nest in else; method already returns early

PleaseInsertNameHere (changes requested):

PleaseInsertNameHere (approved):

tunikakeks (approved):

PR #2889, opened 2026-07-27
fix: add mooshroom variant logic and prevent lightning bolt from striking entities multiple times

iYozemMc:


xRookieFight (changes requested):
All good, but you need to make variant's easy to add a new one. Currently it seems hard to implement a new variant with the ternary operators.

xRookieFight (approved):

PleaseInsertNameHere (approved):

PR #2891, opened 2026-07-27
fix: deadlock in missing BEs

sircurse:
Fix deadlock in missing BEs
Similar to #2875


sircurse:
Did not see the #2890

PR #2911, opened 2026-07-28
perf: optimize PlayerChunkManager

ayzrix-mcbe:
This PR optimizes hot paths in PlayerChunkManager, which runs every tick for every connected player. The comparator context (yaw/FOV trig) is now only recomputed when the player changes chunk instead of every tick, and the FOV check itself avoids Math.sqrt by comparing squared values. The priority queues switch from LongArrayPriorityQueue (O(n) insertion) to LongHeapPriorityQueue (O(log n)), since they're fully rebuilt on each chunk change. A few smaller redundant computations were also removed, including a pointless timed get() call that's already gated behind isDone().

No functional behavior change is expected — these are internal implementation optimizations only.


PleaseInsertNameHere:
Sorry, but this PR violates our contributing guidelines. Please read CONTRIBUTING.md and open again, using our pull request template.

PR #2914, opened 2026-07-28
perf: cache serialised chunk payloads

PleaseInsertNameHere:
## What does this PR do?

- Introduced a new cache: Serialized `ByteBuf` is kept per chunk in the LevelDB provider in LRU order
- Introduced new `IChunk#getContentVersion()` - bumps on `setChanged()` (was unusable since `getChanged()` is zeroed on every save, so an old value could recur and make a stale payload look current (cost: saving no longer invalidates the cache)
- New config var `chunk-settings.payloadCacheMemoryMb` to set max. cache size, `0` to disable
- Also contains a bug fix for block-entity spawn packets (moved out of the serialiser). Went through addChunkPacket, broadcasting to every player holding the chunk, so one player streaming a chunk re-sent every block entity to everyone already standing there (now just sends to the recipient via spawnTo)

Note: Cache is bypassed when anti-xray is enabled.

## Why?

`requestChunkData` re-serialised every chunk from sections, biomes, border columns and block-entity NBT on every request.

## Type of change

- [X] Bug Fix
- [ ] New Feature
- [X] Performance
- [ ] Documentation

## How was this tested?

Live test pending; drafting till completed

None

## Performance notes

N/A

## AI usage disclosure

| Model | What it did |
|-------|-------------|
| Claude Opus 5 | Research based on internal benchmarks, drafted change, javadoc |

## Checklist


xRookieFight (approved):

src/main/java/org/powernukkitx/level/format/leveldb/LevelDBProvider.java
```diff
@@ -378,7 +412,60 @@ public Pair<ByteBuf, Integer> requestChunkData(int x, int z) {
...
+            discardCached(this.payloadCache.put(index,
+                    new CachedChunkPayload(payload, subChunkCount, serializedVersion.get(), payloadBytes)));
+            this.payloadCacheBytes += payloadBytes;
+            while (this.payloadCacheBytes > cacheLimitBytes && !this.payloadCache.isEmpty()) {
+                discardCached(this.payloadCache.removeFirst());
+            }
+        }
+        return Pair.of(payload.retainedDuplicate(), subChunkCount);
+    }
+
```
xRookieFight, line 462: When a single payload is larger than the whole budget, the loop drains the map, including the entry inserted a line earlier. Dropping payload to refCnt: 0. The retainedDuplicate() after the block then throws. I ran your exact insert-and-evict block against real fastutil/netty:

refCnt after serialise     = 1
refCnt after eviction loop = 0
THROWS: io.netty.util.IllegalReferenceCountException: refCnt: 0, increment: 1

subTick catches and logs it, so no crash but processChunkRequest aborts for that tick and will keep aborting every tick, so chunk delivery stalls for anyone whose chunk trips it. Reachable whenever an admin sets payloadCacheMemoryMb low relative to a heavy chunk. Fix is one line: take the caller's duplicate before the synchronized block, or skip caching outright when payloadBytes > cacheLimitBytes.

xRookieFight (commented):
Some hits my AI made:

PooledByteBufAllocator.DEFAULT.ioBuffer() allocates direct buffers, and the cache holds them indefinitely. Nothing in the repo sets -XX:MaxDirectMemorySize or io.netty.maxDirectMemory, so this adds up to 64 MB of pinned off-heap memory and pins pool arenas that would otherwise be recycled — on every existing server, without anyone opting in. For a first release I'd default payloadCacheMemoryMb to 0 and let operators turn it on, or at minimum say in the config comment that this is direct memory, not heap.

Long2ObjectLinkedOpenHashMap.put does not move an existing key to the end. Verified:

initial      = {1, 2, 3}
after re-put = {1, 2, 3}  (first=1)
after touch2 = {1, 3, 2}

A miss only happens when a chunk's version changed, so the chunks that get re-put most often are the busiest ones and they stay pinned at the eviction head and get dropped first. putAndMoveToLast fixes it

getPayloadCacheLimitBytes() is read per request specifically so it can be changed on a live server, but the budget is only enforced on insert. Setting it to 0 makes cacheable false, so no further inserts happen and the existing entries are never released, the memory stays held until unload or shutdown. Worth either enforcing the new budget on read, or dropping the "can be modified on a running server" claim from the javadoc.

Notes, not defects

The version snapshot races with block entities, but fails safe. serializedVersion is read at the top of the batchProcess lambda, and batchProcess holds the block/height/light locks which don't guard tiles. So a block entity added mid-serialisation lands in the payload while the recorded version is one behind. That means the next request sees current != cached and re-serialises: extra work, never a stale send. Worth a comment saying so, since it looks like a bug on first read.

sircurse:
can we postpone merging this one too until LevelDB parity is finished?
I see some concepts here that we can reuse, but parts of this PR are also addressing issues introduced by the current chunk streaming implementation, which levelDB parity is already replacing
once the parity work is finished, I can port the useful concepts over, the new subchunk system will also support caching, but in a quite different way matching BDS

PleaseInsertNameHere:
After your changes have been merged, I'll revise the implementation as well. I see further room for improvement.

PR #2920, opened 2026-07-28
perf: fix async pool thread leak and reduce compression overhead

SemihAkin685:


src/main/java/org/powernukkitx/scheduler/AsyncPool.java
```diff
@@ -16,7 +16,7 @@ public class AsyncPool extends ThreadPoolExecutor {
     private static final AtomicInteger THREAD_COUNTER = new AtomicInteger();
 
     public AsyncPool(int size) {
-        super(size, Integer.MAX_VALUE, 30, TimeUnit.SECONDS, new SynchronousQueue<>());
+        super(size, size, 60, TimeUnit.SECONDS, new LinkedBlockingQueue<>());
```
xRookieFight: With corePoolSize == maximumPoolSize and an unbounded queue, maximumPoolSize can never be reached, the pool is now hard-capped at size threads, that seems off actually


```suggestion
        super(size, size * 4, 60, TimeUnit.SECONDS, new LinkedBlockingQueue<>(1024));
setRejectedExecutionHandler(new CallerRunsPolicy());
```

src/main/java/org/powernukkitx/utils/collection/LZ4Freezer.java
```diff
@@ -7,6 +7,6 @@
 public final class LZ4Freezer {
     public static final LZ4Factory factory = LZ4Factory.fastestInstance();
     public static final LZ4Compressor compressor = factory.fastCompressor();
-    public static final LZ4Compressor deepCompressor = factory.highCompressor();
+    public static final LZ4Compressor deepCompressor = factory.fastCompressor();
```
xRookieFight: deepCompressor is now identical to compressor, which means DEEP_FREEZE and FREEZE produce the same bytes. The entire two-tier temperature system collapses into one tier, and the memory saving that deep freeze exists for is gone. That may well be the right call but highCompressor is roughly an order of magnitude slower. However it's a memory/CPU tradeoff, not a pure perf win.

Two things would help: some before/after numbers, and if you keep it, please go all the way. drop deepCompressor and the DEEP_FREEZE tier rather than leaving code that pretends to do something it no longer does. There's also already a maxCompressionTime budget in FreezableArrayManager.java:182 that could bound HC cost instead, if the goal is just to keep the cycle short.

My personal opinion is reverting this change. If this belongs, there must be a reason. I believe that this breaks some stuff around LZ4 compressing.

src/main/java/org/powernukkitx/utils/collection/FreezableByteArray.java
```diff
@@ -61,11 +61,13 @@ public void freeze() {
     @Override
     public void deepFreeze() {
         if (temperature > manager.getAbsoluteZero()) return;
-        if (freezeStatus.get() != FreezeStatus.NONE && freezeStatus.get() != FreezeStatus.FREEZE) return;
-        var needDecompressFirst = freezeStatus.get() == FreezeStatus.FREEZE;
+        if (freezeStatus.get() == FreezeStatus.FREEZE) {
+            freezeStatus.set(FreezeStatus.DEEP_FREEZE);
+            return;
+        }
```
xRookieFight: Mentioned similar review on AsyncPool, skipping the decompress/recompress is the right instinct. But this relabels fast-compressed data as DEEP_FREEZE, which is only safe because deepCompressor became fastCompressor in this PR. Nothing at this call site says so. Anyone restoring highCompressor later gets an array whose status lies about its contents. Please add a short comment tying the two together, or fold the tiers together properly.

src/main/java/org/powernukkitx/level/format/leveldb/LevelDBProvider.java
```diff
@@ -310,7 +310,7 @@ public Pair<ByteBuf, Integer> requestChunkData(int x, int z) {
         AtomicReference<ByteBuf> data = new AtomicReference<>();
         AtomicReference<Integer> subChunkCountRef = new AtomicReference<>();
         chunk.batchProcess(unsafeChunk -> {
-            final var byteBuf = PooledByteBufAllocator.DEFAULT.ioBuffer();
+            final var byteBuf = PooledByteBufAllocator.DEFAULT.ioBuffer(64 * 1024);
```
xRookieFight: Netty's default is 256 bytes, so this genuinely saves the resize/copy chain for chunk payloads. Only nit: pull 64 * 1024 into a named constant so the intent is visible.

xRookieFight (changes requested):

xRookieFight (approved):
LGTM

xRookieFight:
PR seems off, and violating the guidelines. Sorry!

PR #2924, opened 2026-07-29
fix: prevent chunk save race causing container item loss

SemihAkin685:


src/main/java/org/powernukkitx/level/format/leveldb/LevelDBProvider.java
```diff
@@ -559,6 +561,50 @@ public void saveChunks(Collection<IChunk> chunks) {
...
+                future.complete(collectAndSnapshotDirtyChunks(chunks));
+            } catch (Throwable t) {
+                future.completeExceptionally(t);
+            }
+        });
+        try {
+            return future.get(30, TimeUnit.SECONDS);
+        } catch (Exception e) {
+            return collectAndSnapshotDirtyChunks(chunks);
+        }
```
xRookieFight, line 607: Two things go wrong on timeout:

- The scheduled task is never cancelled. It still runs on a later tick, calling saveNBT(), overwriting serializationSnapshot and clearing setChanged(false) a second time — while the compute pool is concurrently reading those exact fields in LevelDBChunkSerializer.serializeTileAndEntity. That second setChanged(false) can drop every change made between the fallback and the late task. That is precisely the item-loss symptom in the title, just moved.
- The fallback itself runs the snapshot off the main thread, i.e. the unsafe path this PR exists to remove and there's no log line, so nobody would ever know it fired.

Please make it single-winner (future.cancel(true), or an AtomicBoolean both sides claim) and log at WARN. Honestly the safest behaviour on timeout is to skip the save this cycle entirely so the chunks stay dirty and the next autosave picks them up. A delayed save is much cheaper than a corrupted one.

src/main/java/org/powernukkitx/level/format/leveldb/LevelDBProvider.java
```diff
@@ -559,6 +561,50 @@ public void saveChunks(Collection<IChunk> chunks) {
...
+        CompletableFuture<Collection<IChunk>> future = new CompletableFuture<>();
+        server.getScheduler().scheduleTask(null, () -> {
+            try {
+                future.complete(collectAndSnapshotDirtyChunks(chunks));
+            } catch (Throwable t) {
+                future.completeExceptionally(t);
+            }
+        });
+        try {
+            return future.get(30, TimeUnit.SECONDS);
```
xRookieFight: saveChunks is reached from doAutoSave, which runs on the ForkJoin common pool (Server.java:1214). So every autosave now blocks a common-pool thread for at least a full tick, and up to 30s if the main thread lags or the scheduler stops ticking during shutdown. There's also no guard against overlapping autosaves, so a stretch of slow ticks can stack them.

A budget of a couple of ticks (a few hundred ms) plus the skip-this-cycle behaviour above would be much safer than 30s.

src/main/java/org/powernukkitx/level/format/leveldb/LevelDBProvider.java
```diff
@@ -559,6 +561,50 @@ public void saveChunks(Collection<IChunk> chunks) {
         }
     }
 
+    private Collection<IChunk> snapshotDirtyChunks(Collection<IChunk> chunks) {
```
xRookieFight: snapshotDirtyChunks / collectAndSnapshotDirtyChunks aren't snapshots actually, they clear setChanged(false) and rewrite serializationSnapshot. Something like beginSave or claimDirtyChunks would read a lot more honestly about the side effects.

src/main/java/org/powernukkitx/level/format/leveldb/LevelDBProvider.java
```diff
@@ -559,6 +561,50 @@ public void saveChunks(Collection<IChunk> chunks) {
...
+        Server server = Server.getInstance();
+        if (server == null || server.isPrimaryThread()) {
+            return collectAndSnapshotDirtyChunks(chunks);
+        }
```
xRookieFight, line 574: Server doesn't returns null, except if you are running tests. Please comment that

src/main/java/org/powernukkitx/level/format/leveldb/LevelDBProvider.java
```diff
@@ -559,6 +561,50 @@ public void saveChunks(Collection<IChunk> chunks) {
...
+        CompletableFuture<Collection<IChunk>> future = new CompletableFuture<>();
+        server.getScheduler().scheduleTask(null, () -> {
```
xRookieFight: There's a plain scheduleTask(Runnable) overload. The null plugin works but it reads like a mistake.

src/main/java/org/powernukkitx/level/Level.java
```diff
@@ -5294,20 +5294,6 @@ private int unloadChunksAndCount(int maxUnload, boolean force) {
...
-                            }
-                        }
-                        for (Entity e : c.getEntities().values()) {
-                            if (!(e instanceof Player) && !e.closed) {
-                                e.saveNBT();
-                                e.serializationSnapshot = e.getNbt().copy();
-                            }
-                        }
-                    }
                     requireProvider().saveChunks(chunksToSave);
```
xRookieFight, line 5300: unloadChunk still snapshots inline and then calls saveChunk(x, z), which bypasses the new path entirely. It's not incorrect, it's on the main thread but we now have two mechanisms doing the same job. Worth either routing it through the new helper or leaving a comment saying why it stays.

xRookieFight (changes requested):
Also on LevelDBChunkSerializer.java:356, the race is not really closed. 'serializeTileAndEntity' still copies chunk.getBlockEntities().values() / getEntities().values() on a compute thread while the main thread can add and remove entries, and it still falls back to saveNBT() + getNbt().copy() off-thread whenever serializationSnapshot == null. Pre-existing, not caused by you but since the PR title claims to prevent the save race, it's worth stating in the description which part is covered and which isn't.

xRookieFight:
Thank you for taking the time to contribute.

Unfortunately, we're going to close this pull request. Throughout the review process, we decided that this PR is not really fixing the issue, making it even worse.

We appreciate your interest in contributing. If you'd like to submit another pull request in the future, please make sure to carefully test the PR, follow the project's conventions, and ensure that the PR description is fully understood, reviewed, and adapted before submission.

PR #2925, opened 2026-07-29
Add village raid system

YnXahgkiO:
Adds the village raid system (pillager waves scaled by difficulty, boss bar, Bad Omen / Raid Omen / Village Hero effects).


Buddelbubi:
Does not integrate int PNX village system.

PR #2926, opened 2026-07-29
Add village raid system

YnXahgkiO:
Adds the village raid system (pillager waves scaled by difficulty, boss bar, Bad Omen / Raid Omen / Village Hero effects).

Replaces #2925, which used ad-hoc entity scanning instead of the village system. This version detects and targets raids through VillageManager/Village (getVillageAt, isValid, dwellers) instead of scanning nearby entities, and publishes live raid state onto Village.raid().

### DIFF
--- a/src/main/java/org/powernukkitx/Player.java
+++ b/src/main/java/org/powernukkitx/Player.java
@@ -3615,6 +3615,17 @@ public class Player extends EntityHuman implements CommandSender, ChunkLoader, I
         this.sendPacket(pk);
     }
 
+    public void sendTranslatedTitle(String key, int fadeIn, int stay, int fadeOut) {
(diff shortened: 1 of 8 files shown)


Buddelbubi:
It still adds a Raid class over using VillageRaid.
Also why do you need a RaidManager. Cannot that be done in VillageManager?

In case you use AI, please use it correctly. AI tends to build on top of code instead of integrating in existing code because its more work to check if something broke.

YnXahgkiO:
You're right. I'll rework it to fit the existing architecture instead of adding extra abstractions. I'll reopen this PR once I have time to rework the village raids. I tested it successfully on the migration version, but I need to retry everything on PNX 3.0.0.

PR #2950, opened 2026-08-03
fix: spawning two block entities when direct is true

Buddelbubi:


PleaseInsertNameHere (approved):

sircurse (approved):

xRookieFight (approved):

src/main/java/org/powernukkitx/level/Level.java
```diff
@@ -1788,7 +1788,7 @@ public void sendBlocks(Player[] target, Vector3[] blocks, Set<UpdateBlockPacket.
             int runtimeId;
             if (b instanceof Block block) {
                 runtimeId = block.getRuntimeId();
-                if (block instanceof BlockEntityHolder<?> holder && holder.getOrCreateBlockEntity() instanceof BlockEntitySpawnable spawnable) {
+                if (block instanceof BlockEntityHolder<?> holder && holder.getBlockEntity() != null && holder.getBlockEntity() instanceof BlockEntitySpawnable spawnable) {
```
xRookieFight: instanceof is already null-safe. It evaluates to false for null. The first clause can never change the outcome.

Also getBlockEntity() is called twice per block, which isn't good practice

xRookieFight:
```suggestion
                if (block instanceof BlockEntityHolder<?> holder && holder.getBlockEntity() instanceof BlockEntitySpawnable spawnable) {
```

xRookieFight (approved):
LGTM. Just some minor nits i have found

PR #2959, opened 2026-08-04
feat: add /transfer command

iYozemMc:
## What does this PR do?

add /transfer command

## Why?

For add /transfer

- [ ] Bug Fix
- [x] New Feature
- [ ] Performance
- [ ] Documentation

## How was this tested?

- **Server build tested:** 9ae42b2a9
- **Bedrock client version:** 26.40 and 26.44
- **Plugins loaded:** none

1. join server
2. test /transfer on self and another player

None

## Performance notes
N/A

## AI usage disclosure

## Checklist


iYozemMc:
I'm not proud of how the reason is handled, but BedrockDisconnectReasons isn't used anywhere

src/main/java/org/powernukkitx/command/defaults/TransferServerCommand.java
```diff
@@ -0,0 +1,108 @@
...
+                }
+
+                player.transferring = true;
+                player.transfer(address);
+                log.addSuccess("commands.transferserver.successful");
+                log.output();
+                return 1;
+            }
+            case "target" -> {
+                List<Player> targets = list.getResult(0);
```
xRookieFight, line 68: CommandParameter.newType("player", CommandParamType.SELECTION) is declared without a node. In ParamTree.java:65 the default node for SELECTION is EntitiesNode, so the result is actually a List<Entity>. Line 70's List<Player> targets = list.getResult(0) slips through silently thanks to erasure, but the moment targets.getFirst() on line 76 is assigned to a Player, /transferserver @e 1.2.3.4 19132 - or any selector that matches a mob - takes the server down with a ClassCastException.

src/main/java/org/powernukkitx/command/defaults/TransferServerCommand.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Override
+    public int execute(CommandSender sender, String commandLabel, Map.Entry<String, ParamList> result, CommandLogger log) {
+        String overload = result.getKey();
+        ParamList list = result.getValue();
+
+        switch (overload) {
+            case "self" -> {
+                if (!(sender instanceof Player player)) {
+                    log.addMessage("commands.generic.ingame");
```
xRookieFight: There is no commands.generic.ingame key. In eng/lang.json:1115 the key is nukkit.command.generic.ingame (and SimpleCommand.java:82 uses that one). As it stands, running /transferserver <ip> <port> from console prints the raw untranslated key to the sender. commands.generic.player.notFound on line 72 is correct, by the way - only this one is wrong.

src/main/java/org/powernukkitx/command/defaults/TransferServerCommand.java
```diff
@@ -0,0 +1,108 @@
...
+                return 1;
+            }
+            case "target" -> {
+                List<Player> targets = list.getResult(0);
+                if (targets.isEmpty()) {
+                    log.addMessage("commands.generic.player.notFound");
+                    log.output();
+                    return 0;
+                }
+                Player target = targets.getFirst();
```
xRookieFight: Because of targets.getFirst(), /transferserver @a <ip> <port> transfers a single player and silently drops the rest. Every multi-target command in the project (KickCommand.java:47, TitleCommand, FogCommand, etc.) iterates the whole list and ends with log.successCount(players.size()).output(true). This should do the same - loop, and return the actual count rather than return 1.

src/main/java/org/powernukkitx/command/defaults/TransferServerCommand.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+            default -> {
+                return 0;
+            }
+        }
+    }
+
+    private InetSocketAddress resolveAddress(String ip, int port) {
+        try {
+            InetSocketAddress address = new InetSocketAddress(ip, port);
```
xRookieFight: new InetSocketAddress(ip, port) performs a blocking DNS lookup when given a hostname. This command runs on the tick thread, so an unreachable DNS server can freeze the server for seconds. On top of that, Player.transfer (Player.java:5408) already converts to a raw IP via address.getAddress().getHostAddress() - which breaks vanilla behaviour on proxies that route by hostname; BDS sends the string to the client as-is. The cleanest fix: validate the port with a simple port < 0 || port > 65535 check, pass the IP/hostname through unresolved, and leave resolution to the client.

src/main/java/org/powernukkitx/command/defaults/TransferServerCommand.java
```diff
@@ -0,0 +1,108 @@
...
+                    return 0;
+                }
+                Player target = targets.getFirst();
+
+                String ip = list.getResult(1);
+                int port = list.getResult(2);
+
+                InetSocketAddress address = resolveAddress(ip, port);
+                if (address == null) {
+                    log.addMessage("commands.transferserver.invalid.port");
```
xRookieFight: resolveAddress returns null both for an invalid port and for an unresolvable host, yet the user sees "Invalid port (0-65535)" in both cases. That's confusing when the host is what was mistyped. Splitting these as described in item 4 resolves this on its own.

src/main/java/org/powernukkitx/entity/Entity.java
```diff
@@ -210,6 +210,7 @@ public abstract class Entity extends Location implements Metadatable, EntityID {
     protected int lastPlayerNearbyTick = 0;
     public double highestPosition;
     public boolean closed = false;
+    public boolean transferring = false;
```
xRookieFight: Transfer is a Player-specific concept; adding a boolean to every Entity is both semantically wrong and unnecessary. This field belongs on Player. Also please use Getter & Setter's with lombok

src/main/java/org/powernukkitx/command/defaults/TransferServerCommand.java
```diff
@@ -0,0 +1,108 @@
...
+                    log.output();
+                    return 0;
+                }
+
+                target.transferring = true;
```
xRookieFight: Having the command set the flag from outside (lines 63, 88) instead of setting it inside Player.transfer() isn't right - plugins calling player.transfer() today don't get this behaviour at all, so the same operation yields two different outcomes. And the flag is never reset to false: if the transfer fails client-side and the player stays connected, any later genuine disconnect reason gets masked as "Player transferred".

src/main/java/org/powernukkitx/command/defaults/TransferServerCommand.java
```diff
@@ -0,0 +1,108 @@
+package org.powernukkitx.command.defaults;
+
+import org.powernukkitx.Player;
+import org.powernukkitx.command.CommandSender;
+import org.powernukkitx.command.ConsoleCommandSender;
```
xRookieFight: Unused import

src/main/java/org/powernukkitx/command/SimpleCommandMap.java
```diff
@@ -166,6 +166,7 @@ private void setDefaultCommands() {
         this.register("nukkit", new SetWorldSpawnCommand("setworldspawn"));
         this.register("nukkit", new TeleportCommand("tp"));
         this.register("nukkit", new TimeCommand("time"));
+        this.register("nukkit", new TransferServerCommand("transferserver"));
```
xRookieFight: The registration landed between TimeCommand and TitleCommand; alphabetically it belongs after TitleCommand.

src/main/java/org/powernukkitx/command/defaults/TransferServerCommand.java
```diff
@@ -0,0 +1,108 @@
...
+                log.output();
+                return 1;
+            }
+            default -> {
+                return 0;
+            }
+        }
+    }
+
+    private InetSocketAddress resolveAddress(String ip, int port) {
```
xRookieFight: 'resolveAddress' uses no instance state, so it can be private static.

xRookieFight (changes requested):

TraxTRL:
The vanilla command is just `/transfer`. We should adapt to it imo.

Reference: https://tomocraft.github.io/bedrock-commands-docs/#cmd-transfer

src/main/java/org/powernukkitx/Player.java
```diff
@@ -5399,18 +5400,31 @@ public void setSprinting(boolean value) {
...
     }
 
+    /**
+     * Kept for compatibility / internal calls that already have a resolved address.
+     * Never construct the InetSocketAddress from raw user input:
+     * the String,int constructor performs a blocking DNS lookup, and getHostAddress()
+     * silently replaces the hostname with an IP address, which breaks name-based routing
+     * on the proxy/BDS side.
+     */
+
```
TraxTRL: Delete this empty line please
```suggestion
```

iYozemMc (commented):
x

AzaleeX (approved):

tunikakeks (approved):

src/main/java/org/powernukkitx/command/defaults/TransferCommand.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public int execute(CommandSender sender, String commandLabel, Map.Entry<String, ParamList> result, CommandLogger log) {
+        String overload = result.getKey();
+        ParamList list = result.getValue();
+
+        switch (overload) {
+            case "self" -> {
+                if (!(sender instanceof Player player)) {
+                    log.addMessage("nukkit.command.generic.ingame");
+                    log.output();
```
TraxTRL:
```suggestion
                    log.addMessage("nukkit.command.generic.ingame").output();
```

src/main/java/org/powernukkitx/command/defaults/TransferCommand.java
```diff
@@ -0,0 +1,110 @@
...
+                    return 0;
+                }
+
+                String ip = list.getResult(0);
+                int port = list.getResult(1);
+
+                if (!isValidPort(port)) {
+                    log.addMessage("commands.transfer.invalid.port");
+                    log.output();
```
TraxTRL:
```suggestion
                    log.addMessage("commands.transferserver.invalid.port").output();
```

src/main/java/org/powernukkitx/command/defaults/TransferCommand.java
```diff
@@ -0,0 +1,110 @@
...
+                    return 0;
+                }
+
+                player.transfer(ip, port);
+                log.addSuccess("commands.transfer.successful");
+                log.output();
```
TraxTRL:
```suggestion
                log.addSuccess("commands.transferserver.successful").output(true);
```

src/main/java/org/powernukkitx/command/defaults/TransferCommand.java
```diff
@@ -0,0 +1,110 @@
...
+                return 1;
+            }
+            case "target" -> {
+                List<?> rawTargets = list.getResult(0);
+                if (rawTargets.isEmpty()) {
+                    log.addMessage("commands.generic.player.notFound");
+                    log.output();
```
TraxTRL:
```suggestion
                    log.addMessage("commands.generic.player.notFound").output();
```

src/main/java/org/powernukkitx/command/defaults/TransferCommand.java
```diff
@@ -0,0 +1,110 @@
...
+                for (Object obj : rawTargets) {
+                    if (obj instanceof Player p) {
+                        target = p;
+                        break;
+                    }
+                }
+
+                if (target == null) {
+                    log.addMessage("commands.generic.player.notFound");
+                    log.output();
```
TraxTRL:
```suggestion
                    log.addMessage("commands.generic.player.notFound").output();
```

src/main/java/org/powernukkitx/command/defaults/TransferCommand.java
```diff
@@ -0,0 +1,110 @@
...
+                    return 0;
+                }
+
+                String ip = list.getResult(1);
+                int port = list.getResult(2);
+
+                if (!isValidPort(port)) {
+                    log.addMessage("commands.transfer.invalid.port");
+                    log.output();
```
TraxTRL:
```suggestion
                    log.addMessage("commands.transferserver.invalid.port").output();
```

src/main/java/org/powernukkitx/command/defaults/TransferCommand.java
```diff
@@ -0,0 +1,110 @@
...
+                    return 0;
+                }
+
+                target.transfer(ip, port);
+                log.addSuccess("commands.transfer.successful");
+                log.output();
```
TraxTRL:
```suggestion
                log.addSuccess("commands.transferserver.successful").output(true);
```

src/main/java/org/powernukkitx/command/defaults/TransferCommand.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @author iYozem
+ */
+public class TransferCommand extends VanillaCommand {
+
+    public TransferCommand(String name) {
+        super(name, "commands.transfer.description");
```
TraxTRL:
```suggestion
        super(name, "commands.transferserver.description");
```

src/main/java/org/powernukkitx/command/defaults/TransferCommand.java
```diff
@@ -0,0 +1,110 @@
...
+        this.setPermission("nukkit.command.transferserver");
```
TraxTRL:
```suggestion
        this.setPermission("nukkit.command.transfer");
```
You also need to register the permission in DefaultPermissions class

src/main/java/org/powernukkitx/network/process/NetworkPacketHandler.java
TraxTRL, line 1: Not sure if we really need this. I would prefer sticking with protocol standards. We can broadcast the command output instead.

TraxTRL (changes requested):
Revert the language file changes

Reference: https://github.com/Mojang/bedrock-samples/blob/main/resource_pack/texts/en_US.lang#L1502-L1504

TraxTRL (changes requested):
Add the transfer command permission in DefaultPermissions class please

src/main/java/org/powernukkitx/permission/DefaultPermissions.java
```diff
@@ -106,6 +106,7 @@ public static void registerCorePermissions() {
         //registerPermission(new Permission(ROOT + ".command.dumpmemory", "Allows the user to dump memory contents", Permission.DEFAULT_OP), commands); // this command is not implemented
         registerPermission(new Permission(ROOT + ".command.gamerule", "Sets or queries a game rule value", Permission.DEFAULT_OP), commands);
         registerPermission(new Permission(ROOT + ".command.title", "Allows the user to send titles to players", Permission.DEFAULT_OP), commands);
+        registerPermission(new Permission(ROOT + ".command.transferserver", "Allows the user to transfer a player", Permission.DEFAULT_OP), commands);
```
TraxTRL: In the command you set the permission to `nukkit.command.transfer`, so fix this here please

src/main/java/org/powernukkitx/command/defaults/TransferCommand.java
TraxTRL, line 1: Just a suggestion, but we could split the permission to .self and .other (like for kill command) allowing server operators more customization

TraxTRL (changes requested):

src/main/java/org/powernukkitx/permission/DefaultPermissions.java
```diff
@@ -77,6 +77,11 @@ public static void registerCorePermissions() {
         registerPermission(new Permission(ROOT + ".command.kill.other", "Allows the user to kill other players"), kill);
         kill.recalculatePermissibles();
 
+        Permission transferserver = registerPermission(new Permission(ROOT + ".command.transferserver", "Allows the user to transfer players", Permission.DEFAULT_OP), commands);
+        registerPermission(new Permission(ROOT + ".command.transferserver.self", "Allows the user to transfer themselves to another server", Permission.DEFAULT_TRUE), transferserver);
+        registerPermission(new Permission(ROOT + ".command.transferserver.other", "Allows the user to transfer other players to another server", Permission.DEFAULT_OP), transferserver);
+        transferserver.recalculatePermissibles();
```
TraxTRL: For consistency with other command permissions, rename to just transfer instead of transferserver
Don't forget to also update it in the command class itself too

src/main/java/org/powernukkitx/Player.java
```diff
@@ -270,6 +270,7 @@ public class Player extends EntityHuman implements CommandSender, ChunkLoader, I
     protected int viewDistance;
     protected Position spawnPoint;
     protected SpawnPointType spawnPointType;
+    public boolean transferring = false;
```
TraxTRL: Remove this and its usages

src/main/java/org/powernukkitx/network/process/NetworkPacketHandler.java
TraxTRL, line 1: Revert this; see my other comment on this file for reason

TraxTRL (changes requested):

TraxTRL:
I tested this in-game and applied changes directly because I felt like it is easier that way, I hope you don't mind.

iYozemMc:
np

TraxTRL (approved):
Please another maintainer review before merging this as I pushed to this PR and don't like self-approving

xRookieFight (approved):

PR #2972, opened 2026-08-07
fix(network): harden packet handling against malicious clients

Wyze3306:
## What does this PR do?

Fixes 20 security issues in the network layer and the inbound packet handlers. Every one of them
can be triggered by a normal client, with no operator rights and no plugin involved.

## Why?

I audited the path an inbound packet takes, from the codec down to the handlers, after two of these
problems hit my own production server: the console filled up with the same decode error forever,
and a RakNet flood from a single address was never rate limited.

Reading the rest of that path turned up more, including an item duplication through the recipe book,
a mining loop that never ends server-side, and a path traversal reachable before login. Two of the
findings are dead code rather than hardening: the botnet detector is never called outside the tests,
so it has never detected anything, and `disconnect()` only ever sent a packet without closing the
session, so no rejection actually dropped a connection.

The fixes, grouped:

**Denial of service from client input**
- A packet the codec cannot decode left the session open, so the client kept re-sending it and the
  console filled with identical stack traces. The channel is now closed on the first decode failure,
  and a `BurstFilter` caps the codec logger in case an attacker reconnects and floods again.
- `RAK_GLOBAL_PACKET_LIMIT` was never passed to the channel, so the server-wide backstop against
  distributed floods was left at the library default.
- `breakingBlock` was only ever cleared by a client packet. A client that sends
  `START_DESTROY_BLOCK` and never sends `ABORT` left the tick loop broadcasting a `LevelEventPacket`
  to the whole chunk indefinitely, and recomputing the break time every tick.
- `MapInfoRequestPacket` scanned every block entity in the level, with no rate limit and no distance
  or ownership check, so it also returned the image of any map anywhere on the server.
- `CraftRecipeAutoProcessor` matched ingredients in O(n²); with the codec's list limit that is about
  2.4M iterations with allocations per action.
- Handlers that make the server broadcast on the client's behalf had no limit of their own.

**Duplication**
- The auto craft path multiplied the output by the requested craft count without ever reading the
  inventory, and only counted the consume actions rather than their amounts. The output is now tied
  to the amount actually consumed.

**Permission and authentication**
- `SettingsCommandPacket` executed commands with no rate limit, no state check, and without firing
  `PlayerCommandPreprocessEvent`, so every plugin filtering or logging commands through that hook was
  bypassed. It now follows the same path as `CommandRequestHandler`.
- Handlers run inline on the Netty thread until the player exists, so their code was reachable before
  login. Added a `requiresPlayer()` gate.

**File access**
- The structure name went straight from the client into a `File` path with no validation. Added a
  character allowlist and a canonical containment check, and made the cache concurrent since it is
  written from several threads.

**Missing reach checks**
- Entity interaction, lecterns, NPCs, `onBlockBreakStart` and `CREATIVE_DESTROY_BLOCK` allowed acting
  at an arbitrary distance, and the last two could force chunk generation at arbitrary coordinates.

**Smaller ones**
- The NaN guard on player rotation was reading the position, so rotation was never validated.
- `LevelSoundEventPacket` was relayed verbatim to other clients; it is now rebuilt server-side.
- The villager trade placed its output before validating the amounts.
- `decreaseCount` had no sign guard, so a negative amount would have increased the stack.
- `RateLimiter.create(0)` made player creation fail when a limit was set to zero in the config. Zero
  now means no limit for that category.

Closes #

## Type of change

- [x] Bug Fix
- [ ] New Feature
- [ ] Performance
- [ ] Documentation

## How was this tested?

- **Server build tested:** `2143aaa29` — this is my fork, carrying the same diff as the branch
  (`1b3c34562`), not the branch commit itself. I checked the two are equivalent on the security
  content before testing.
- **Bedrock client version:** 1.26.40
- **Plugins loaded:** none

1. Built the server and ran `gradlew test`: 1002 tests, 0 failures. Started the JAR from a clean
   directory, it reaches the setup wizard with no log4j or config error.
2. Ran the server with no plugins and connected with a Bedrock client. Joined, loaded chunks, walked
   around, no disconnection and nothing unusual in the console.
3. Crafted from the recipe book. Normal crafts still work; the duplication fix only refuses requests
   that claim more output than they consume.
4. Traded with a villager, used a lectern, interacted with entities, played an emote, and made
   sounds, to exercise the new rate limit and the reach checks. Normal play is unaffected.

- [x] Existing unit tests pass (`gradlew test`) — 1002 tests, 0 failures

| Change | Impact |
|---|---|
| `PacketHandler.requiresPlayer()` | New `default` method, so source compatible. But any third-party handler inherits `true`, and its packets are now dropped until the `Player` exists. A handler that is part of the login sequence has to override it to `false`. |
| `PlayerSessionHolder.disconnect(...)` | Now closes the session (`session.close()`). A plugin calling `disconnect()` and expecting the session to stay alive afterwards will see different behaviour. |
| `Inventory.decreaseCount(slot, amount)` | An `amount <= 0` is now ignored instead of **increasing** the stack. |
| `LevelSoundEventPacket` | No longer relayed verbatim: rebuilt server-side, `isGlobal` forced off, `actorUniqueId` set to the player, and the position clamped to 16 blocks from the player. A plugin relying on the raw relay loses those fields. |
| `MapInfoRequestPacket` | An item frame further away than the player's view distance no longer returns its map image, where the scan used to cover the whole level. Maps held in the inventory or the offhand are unchanged. |
| Config | Two keys added: `network.globalPacketLimit` (100000) and `network.rate-limit.maxWorldInteractionPacketsPerSecondPerPlayer` (30). Both are appended to an existing `pnx.yml`; no existing value is changed. |
| Crafting | Auto craft now refuses requests that were accepted before, namely those claiming more output than they consume. |
| Logging | `ItemStackRequestHandler` drops its unhandled-action message to `debug`, since one request can carry many actions and is not rate limited. Other warnings are unchanged. |

## Performance notes

Three changes go the right way and none of them costs anything:

- `CraftRecipeAutoProcessor`: ingredient matching goes from O(n²) to O(n). With the codec's list
  limit of 1536 the worst case was about 2.4M iterations with allocations, per action.
- `MapInfoRequestHandler`: the level-wide block entity scan is now distance filtered and rate
  limited. It is still O(n) in the number of block entities.
- `PlayerAuthInputHandler`: a `PlayerHandle` allocation moved out of a loop that runs up to 32 times
  per packet, 20+ packets per second per player.

## AI usage disclosure

| Model | What it did |
|-------|-------------|
| Claude Opus 5 (`claude-opus-5`) | Edited files across the repository. Audited the inbound packet path, identified the vulnerabilities, wrote the whole diff, the commit message and the Javadoc.|

## Checklist


AzaleeX (approved):

src/main/java/org/powernukkitx/Player.java
```diff
@@ -603,6 +613,29 @@ public void onBlockBreakStart(Vector3 pos, BlockFace face) {
...
+        if (!(miningTimeRequired > 0) || !Double.isFinite(miningTimeRequired)) {
+            return false;
+        }
+        if (miningTimeRequired > MAX_BLOCK_BREAK_SECONDS) {
+            return false;
+        }
+        if (now - this.breakingBlockTime > BLOCK_BREAK_STALL_MS) {
+            return false;
+        }
+        if (this.breakingBlock.getLevel() != this.level) {
```
xRookieFight, line 625: The BLOCK_BREAK_STALL_MS check is dead. onBlockBreakContinue is driven by the tick loop (Player.java:3109), and it refreshes breakingBlockTime = time on every pass (Player.java:527), so now - this.breakingBlockTime is always about one tick. The never-ending mine is actually stopped by the finite/300s/reach checks; the stall guard never fires and reads as if it does.

src/main/java/org/powernukkitx/network/process/PlayerSessionHolder.java
```diff
@@ -155,11 +155,14 @@ public void disconnect(DisconnectFailReason reason) {
...
 
     public void disconnect(DisconnectFailReason reason, String message) {
+        this.disconnected = true;
+
         final DisconnectPacket packet = new DisconnectPacket();
         packet.setReason(reason);
         packet.setMessages(new DisconnectPacketMessages(message, ""));
 
         this.session.sendPacketImmediately(packet);
+        this.session.close(message);
```
xRookieFight, line 165: sendPacketImmediately(DisconnectPacket) then session.close(message). close(String) sends its own disconnect, so the client now receives two. Use the no-arg close() here, or drop the hand-built packet.

src/main/java/org/powernukkitx/network/security/PacketRateLimiter.java
```diff
@@ -16,12 +16,18 @@ public final class PacketRateLimiter {
...
+        this.command = create(settings.maxCommandsPerSecondPerPlayer());
+        this.chat = create(settings.maxChatPerSecondPerPlayer());
+        this.formResponse = create(settings.maxFormResponsesPerSecondPerPlayer());
+        this.movement = create(settings.maxMovementPacketsPerSecondPerPlayer());
+        this.worldInteraction = create(settings.maxWorldInteractionPacketsPerSecondPerPlayer());
+    }
+
+    private static RateLimiter create(int permitsPerSecond) {
+        return RateLimiter.create(Math.max(1, permitsPerSecond));
     }
```
xRookieFight, line 38: Math.max(1, permitsPerSecond) gives a configured 0 a meaning it never had. If 0 was intended as "unlimited" this throttles to 1/s; if it was "block everything" it now leaks one per second. Pick one and say so in the config comment.

src/main/java/org/powernukkitx/network/process/handler/LevelSoundEventHandler.java
```diff
@@ -1,21 +1,54 @@
...
+            return;
         }
+
+        String identifier = packet.getActorIdentifier();
+        if (identifier == null) {
+            identifier = "";
+        } else if (identifier.length() > MAX_IDENTIFIER_LENGTH) {
+            return;
+        }
+
```
xRookieFight, line 43: The rebuilt packet uses the player's position and drops the client's. Sounds that genuinely originate away from the player (bucket fill on a block, note block, armor stand) now play at the player instead. Clamping the client position to a max distance keeps the anti-spoof property without relocating every sound.

src/main/java/org/powernukkitx/network/process/handler/MapInfoRequestHandler.java
```diff
@@ -59,8 +62,16 @@ public void handle(MapInfoRequestPacket packet, PlayerSessionHolder holder, Serv
         }
 
         if (mapItem == null) {
+            int radius = player.getViewDistance() << 4;
+            long radiusSquared = (long) radius * radius;
             for (BlockEntity be : player.level.getBlockEntities().values()) {
-                if (be instanceof BlockEntityItemFrame itemFrame && checkMapItemValid(itemFrame.getItem(), packet)) {
+                if (!(be instanceof BlockEntityItemFrame itemFrame)) {
```
xRookieFight, line 68: The new radius is player.getViewDistance() << 4, so an item frame just outside render distance stops returning its map image. That is a gameplay change, not only hardening; call it out in the breaking-changes table.

xRookieFight (changes requested):
On PacketViolationgWarningHandler:16, it inherits requiresPlayer() == true, but this packet is what a client sends when it cannot parse something the server sent, which is most likely during login or the resource-pack phase, before the player exists. The gate silently discards the only report of server-side protocol bugs. Override to false, and null-guard holder.getPlayer().getName() on that line so it NPEs pre-player today.

On TickSyncHandler:23, same shape: runsOnNetworkThread() == true plus an unguarded holder.getPlayer(). The new gate suppresses the NPE by dropping the packet. Worth confirming the client does not send TickSync before player creation, otherwise tick sync now fails silently instead of loudly.

xRookieFight (commented):
Lots of this is solid: the `getPosition()` to `getPlayerRotation()` fix, the `decreaseCount` sign guard, hoisting `PlayerHandle` out of the block action loop, wiring `RAK_GLOBAL_PACKET_LIMIT`, and the `StructureAPI` allowlist plus canonical containment check. Six notes:

**`CraftRecipeAutoProcessor.java:54`**

The match loop validates nothing. `eventItems` is built on line 45 from `action.getIngredients()` itself, so the loop compares each ingredient descriptor against an item derived from that same descriptor, and `matched` is a tautology. The player's inventory is never consulted. That was already true before this PR, so the O(n²) to O(n) rewrite is a fair perf win, but the description reads as though this were an ingredient check and it isn't. Same for the `CraftItemEvent` on line 47: plugins get the recipe's own ingredients, not what the player holds.

**`CraftRecipeAutoProcessor.java:102`**

The new consume check compares two client supplied numbers. `requiredTotal` comes from `getStackSize()` and `timesCrafted`, `consumedTotal` from `consumeAction.getAmount()`, and both are fields of the same request, so a client wanting more output just claims a bigger consume amount. The real bound lives in `ConsumeActionProcessor` (`item.getCount() < count` errors), which runs after this processor already wrote the output, and `ItemStackRequestHandler.handleRequests` only does `break` on a failed action with no rollback, so the output stays in `CreativeOutputInventory`. As written the check forces the client to lie consistently rather than tying the output to what it owns. Closing it needs either a rollback or reading the source slots here.

**`log4j2.xml:27`**

The `BurstFilter` can't fire. `BedrockPacketCodec` takes its logger from Netty's `InternalLoggerFactory` and logs the decode failure at debug only, guarded by `isDebugEnabled()`, then rethrows. A logger stanza at ERROR with a `BurstFilter` at ERROR filters a stream that doesn't exist. The repeating stack traces come from Netty's pipeline tail (`io.netty.channel.DefaultChannelPipeline`) logging the unhandled `exceptionCaught`, which is exactly what `MalformedPacketGuard` now intercepts. The guard is the real fix; this stanza is dead config.

**`Player.java:483`**

Creative left clicks now emit a stray `BLOCK_STOP_BREAK` broadcast. `onBlockBreakStart` assigns `breakingBlock` in every game mode, and in creative `calculateBreakTime` is 0, so on the next tick `isBreakStillValid` fails on `!(miningTimeRequired > 0)` and calls `onBlockBreakAbort(pos)`, which sends a `LevelEventPacket` to everyone in the chunk. Previously the `if (miningTimeRequired > 0)` block was just skipped. Clearing the state is right, but `onBlockBreakAbort` is the wrong tool for "there was never a break to abort"; `resetBlockBreak()` gives the same cleanup without the per click broadcast.

**`PlayerSessionHolder.java:165`**

`BedrockSession.close(reason)` goes straight to `peer.close(reason)`. The library's own `BedrockServerSession.disconnect(reason, hideReason)` exists for this: it sends the `DisconnectPacket`, blackholes inbound traffic, and schedules the close on the event loop so the client actually receives the reason. Closing the peer immediately after `sendPacketImmediately` risks the player seeing a generic connection loss instead of the kick message. Also, `close(...)` starts with `checkForClosed()`, which throws `IllegalStateException("Session has been closed")`, and `this.disconnected` is now set at the top of both overloads but never read, so a second `disconnect()` throws. An early `if (this.disconnected) return;` would use the field that was just added.

**`Player.java:1194`**

The new catch clause will hide real bugs. `IllegalArgumentException`, `IllegalStateException` and `IndexOutOfBoundsException` are what server side logic errors throw, not only malformed input, so demoting all three to debug across every inbound packet makes a genuine regression in any handler invisible in production. Narrowing it to the handlers that actually parse client input, or keeping a rate limited warn, would keep the protection without the blind spot.

tunikakeks:
@Wyze3306 pls check & resolve conflicts

PR #2981, opened 2026-08-10
fix: prevent DoS through text and settings packets

iYozemMc:
## What does this PR do?

To prevent DoS with specials chars and so that the player doesn't freeze when receiving a barrage of characters

## Why?

security
## Type of change

- [X] Bug Fix
- [ ] New Feature
- [ ] Performance
- [ ] Documentation

## How was this tested?

- **Server build tested:** d16947437
- **Plugins loaded:** none

1. join a server
2. connect a bot with gophertunnel with script

None

## Performance notes

N/A

## AI usage disclosure

| Model | What it did |
|-------|-------------|
|   Claude Fable 5    |     for the sanitizeChatMessage function        |

## Checklist

### DIFF
--- a/src/main/java/org/powernukkitx/network/process/handler/TextHandler.java
+++ b/src/main/java/org/powernukkitx/network/process/handler/TextHandler.java
@@ -15,6 +15,7 @@ import org.cloudburstmc.protocol.bedrock.packet.TextPacket;
  */
 @Slf4j
 public class TextHandler implements PacketHandler<TextPacket> {
+    private static final int MAX_CHAT_LENGTH = 256;
 
     @Override
     public void handle(TextPacket packet, PlayerSessionHolder holder, Server server) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/powernukkitx/network/process/handler/TextHandler.java
TraxTRL, line 1: In my opinion, character length limit and character blocking shouldn't be handled by PNX but a plugin.

iYozemMc: I've thought that before, but I figure it's always better to fix a security vulnerability than to have it done through the software.

xRookieFight: No, we shouldn't handle special character blocking from PNX

PR #2982, opened 2026-08-10
fix: players invisible after re-entering spawn or ticking-area chunks

SemihAkin685:


Buddelbubi:
Why does unregisterChunkLoader return false?

SemihAkin685:
> Why does unregisterChunkLoader return false?


Good catch that was the actual root cause.

The return value conflated two things: whether the loader was removed, and the result of `unloadChunkRequest(safe = true)`, which is false for spawn chunks and ticking areas. So unregistering from a spawn chunk returned false even though the loader was removed, and `unloadChunkForPlayer` skipped its despawn loop no `RemoveActor`, stale `Entity#hasSpawned`, so `spawnTo` no-ops on re-entry and the player stays invisible.

All callers only care about the first meaning, so the second commit makes the unload request a side effect instead of the return value.

xRookieFight:
Closing this PR. There is no description regarding to anything. Also there is undisclosed AI usage. Please fill the PR description and read the PR policy next time.

PR #2984, opened 2026-08-11
fix: client disconnecting on item entities count bigger than 64

xRookieFight:


tunikakeks (approved):

Buddelbubi (approved):

src/main/java/org/powernukkitx/level/Level.java
```diff
@@ -3328,42 +3355,40 @@ public void dropItem(Vector3 source, Item item, Vector3 motion, boolean dropArou
         return this.dropAndGetItem(source, item, motion, false, delay);
     }
 
+    /**
+     * @return the dropped item entity, or the first one if the stack had to be split into several entities
+     */
     public @Nullable EntityItem dropAndGetItem(@NotNull Vector3 source, @NotNull Item item, @Nullable Vector3 motion, boolean dropAround, int delay) {
```
Buddelbubi: You could change this so it returns an array

PleaseInsertNameHere (approved):

src/main/java/org/powernukkitx/level/Level.java
xRookieFight: Why?

Buddelbubi: Because right now it only returns the first dropped entrity<. If split up, the other ones are not returned

PR #2985, opened 2026-08-11
fix: inventory transaction check

iYozemMc:
## What does this PR do?

for security and prevent NPE 

## Why?

security

## Type of change

- [X] Bug Fix
- [ ] New Feature
- [ ] Performance
- [ ] Documentation

## How was this tested?

- **Server build tested:** 5044bcf
- **Plugins loaded:** none

1. with test

- [X] Existing unit tests pass (`gradlew test`) - **I don't want to add another test, knowing that there are already so many.**

None

## Performance notes

N/A

## AI usage disclosure

| Model | What it did |
|-------|-------------|
|       |             |

## Checklist

### DIFF
--- a/src/main/java/org/powernukkitx/network/process/handler/InventoryTransactionHandler.java
+++ b/src/main/java/org/powernukkitx/network/process/handler/InventoryTransactionHandler.java
@@ -70,14 +70,19 @@ public class InventoryTransactionHandler implements PacketHandler<InventoryTrans
             return;
         }
 
-        if (packet.getTransaction().getType().equals(InventoryTransactionDataType.ITEM_USE)) {
-            handleUseItem(playerHandle, (ItemUseInventoryTransaction) packet.getTransaction());
-        } else if (packet.getTransaction().getType().equals(InventoryTransactionDataType.ITEM_USE_ON_ACTOR)) {
-            handleUseItemOnEntity(playerHandle, (ItemUseOnActorInventoryTransaction) packet.getTransaction());
-        } else if (packet.getTransaction().getType().equals(InventoryTransactionDataType.ITEM_RELEASE)) {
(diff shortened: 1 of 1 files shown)


AzaleeX (approved):

src/main/java/org/powernukkitx/network/process/handler/InventoryTransactionHandler.java
```diff
@@ -66,6 +66,10 @@ public class InventoryTransactionHandler implements PacketHandler<InventoryTrans
     public void handle(InventoryTransactionPacket packet, PlayerSessionHolder holder, Server server) {
         final PlayerHandle playerHandle = holder.getPlayerHandle();
         Player player = playerHandle.player;
+        if (!player.spawned || !player.isAlive()) {
```
xRookieFight, line 69: packet.getTransaction() is still dereferenced unguarded, four times in this method and again at lines 105-113. For a PR whose stated goal is NPE prevention, that is the null that a malformed packet actually reaches; the player state was never the wire-controlled part.

src/main/java/org/powernukkitx/network/process/handler/InventoryTransactionHandler.java
xRookieFight, line 400: With the new guard at the top of handle, !player.spawned || !player.isAlive() here is unreachable and only player.isCreative() still does anything. Reduce it, or the next reader will assume DESTROY can be hit while dead.

xRookieFight (changes requested):
Also for the PR description, the quoted failures name InventoryTransactionHandlerTest, which does not exist on beta or in this diff. Either the test file is missing from the PR or the output came from a local branch; worth asking the author which, since the stated verification cannot be reproduced as written.

PR #2988, opened 2026-08-12
feat: added persistent End dragon fight with portal and spike regeneration

NaySurGithub:
## What does this PR do?

This PR add the End dragon fight system.

## Why?

Before this, the main End island and the dragon fight was incomplete.

## Type of change

- [ ] Bug Fix

- [x] New Feature

- [ ] Performance

- [ ] Documentation

## How was this tested?

Regenerated the end world, did a fight against the dragon, tested the crystal respawn system (placed 4 crystal in the edrock portal)

- **Server build tested:** 5605bc9a9

- **Plugins loaded:** none

Added Level#getEnderDragonFight().

## Performance notes

N/A

## AI usage disclosure

| Model | What it did |

|-------|-------------|

|   GPT-5.6-SOL    |    did more similar to minecraft maths        |

- [x] This PR description and any review replies are written by me, not generated

## Checklist

### DIFF
--- a/src/main/java/org/powernukkitx/entity/item/EntityEnderCrystal.java
+++ b/src/main/java/org/powernukkitx/entity/item/EntityEnderCrystal.java
@@ -16,6 +16,8 @@ import org.powernukkitx.nbt.tag.CompoundTag;
 import org.cloudburstmc.protocol.bedrock.data.actor.ActorDataTypes;
 import org.cloudburstmc.protocol.bedrock.data.actor.ActorFlags;
 import org.jetbrains.annotations.NotNull;
+import lombok.Getter;
+import lombok.Setter;
 
 /**
  * @author PetteriM1
@@ -33,6 +35,14 @@ public class EntityEnderCrystal extends Entity implements EntityExplosive {
      */
     protected boolean detonated = false;
 
+    /**
+     * Whether this crystal belongs to an active End dragon respawn sequence.
+     * While {@code true} the crystal is invulnerable.
+     */
+    @Getter
+    @Setter
+    private boolean respawning;
+
     public EntityEnderCrystal(IChunk chunk, CompoundTag nbt) {
         super(chunk, nbt);
     }
@@ -46,6 +56,7 @@ public class EntityEnderCrystal extends Entity implements EntityExplosive {
         if (nbtMap.contains("ShowBottom")) {
             this.setShowBase(nbtMap.getBoolean("ShowBottom"));
         }
+        respawning = nbtMap.getBoolean("Respawning");
     }
 
     @Override
@@ -53,6 +64,7 @@ public class EntityEnderCrystal extends Entity implements EntityExplosive {
         super.saveNBT();
 
         this.nbt.putBoolean("ShowBottom", this.showBase());
+        this.nbt.putBoolean("Respawning", respawning);
     }
 
     @Override
@@ -77,7 +89,7 @@ public class EntityEnderCrystal extends Entity implements EntityExplosive {
 
     @Override
     public boolean attack(EntityDamageEvent source) {
-        if (isClosed()) {
+        if (isClosed() || respawning) {
             return false;
         }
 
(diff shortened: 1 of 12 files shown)


AzaleeX (approved):

src/main/java/org/powernukkitx/entity/item/EntityEnderCrystal.java
```diff
@@ -32,6 +32,7 @@ public String getIdentifier() {
      * @since 1.2.1.0-PN
      */
     protected boolean detonated = false;
+    private boolean respawning;
```
xRookieFight, line 44: We can use lombok setter & getter

xRookieFight (changes requested):

src/main/java/org/powernukkitx/entity/item/EntityEnderCrystal.java
NaySurGithub: resolved

NaySurGithub:
Need a review..

xRookieFight (commented):
Shape of this is good. Moving the podium and gateway logic out of `EntityEnderDragon.kill()` into a level owned controller, and giving `ObjectObsidianPillar` explicit constructor args instead of the mutable `i`/`seed` pair, are both clear wins. Six notes:

**`EnderDragonFight.java:310`**

`scanState()` can never see an already beaten End. It walks the column at `x=0, z=0` looking for a `BlockEndPortal`, but `ObjectExitPortal` writes `BEDROCK` over `(originX, originY..originY+3, originZ)` after laying the portal ring, so that column is always bedrock and `activePortal` is permanently false. Both consequences land on the migration path the scan exists for: a world where players already killed the dragon comes up with `previouslyKilled = false`, so line 313's `if (!previouslyKilled && dragonKilled) dragonKilled = false` respawns the dragon for free and the next kill drops a second egg; and line 322's `if (!activePortal && dragon == null) spawnPodium(false)` fires and rebuilds the podium over whatever sits at origin. Scanning a ring position like `(2, y, 0)` instead of the centre would cover both.

**`EnderDragonFight.java:96`**

`dragonUniqueId == null || ++ticksSinceDragonSeen >= DRAGON_SCAN_INTERVAL` never evaluates the `++` while the id is null, so `findDragon()` walks all of `level.getEntities()` and `createDragon()` runs every tick until an id is set. Usually that resolves in one tick, but `createDragon` only assigns the id inside `if (entity instanceof EntityEnderDragon)`, so if `Entity.createEntity` returns null the id stays null and the level tick keeps trying to spawn a dragon, once per tick, forever.

**`EnderDragonFight.java:186`**

`resetSpike` is a verbatim copy of `ObsidianPillarPopulator.createSpikes`: same seed derivation, same shuffle, same trig. Two copies of the spike table will drift, and when they do the respawned spikes stop matching the generated ones. It also rebuilds the whole 10 entry `Random` and shuffle on each of the ten calls. Exposing the table as a static from the populator and indexing into it removes both. Minor related thing: `resetSpike` passes `level.getHeightMap(x, z)` as the Y, which `ObjectObsidianPillar.generate` now ignores entirely.

**`ObjectObsidianPillar.java:35`**

The loop starts at `getMinHeight()` instead of `0`, which extends every pillar 64 blocks further down into the void. The new `else if (i > 65)` on line 40 also clears air over the full `(2r+1)²` square rather than the circle, up to `height + 10`. Harmless during world generation, but `resetSpike` runs it on a live world, so a respawn deletes anything players built in that square around a spike.

**`EnderDragonFight.java:80`**

The respawn sequence freezes if the player walks away. `tick()` returns early when nobody is within 192 blocks of `(0, 128, 0)`, so `respawnTicks` stops advancing mid sequence while the crystals keep `respawning = true`, which `EntityEnderCrystal.attack` treats as invulnerable. They sit there unbreakable until someone comes back. A timeout, or letting the sequence run to completion regardless of players, would avoid a stuck End.

**`EntityEnderCrystal.java:171`**

`clearBeamTarget()` writes a null into the actor data map, where every other `BLOCK_TARGET` write passes a `Vector3i`, and that value goes straight out through `setDataProperty`. If the codec can't encode null for that field this throws on the send path. It's only reachable from `abortRespawn()`, a crystal destroyed mid sequence, which is easy to miss in manual testing, so worth confirming rather than assuming. Sending `(0, 0, 0)` is the usual way to clear it.

sircurse:
The ender dragon fight has several elements directly linked to leveldb, so I would not move forward with this PR by now, as that is being refactored to match BDS directly (leveldb parity that is coming soon).
We really appreciate the effort.

NaySurGithub:
Being implemented by #3235

PR #3003, opened 2026-08-14
fix: Non-named voxel shapes registry

sircurse:
## What does this PR do?

Fix the registry of non-named voxel shapes that BDS sends to client.

## Why?

Packet parity

Closes #

## Type of change

- [X] Bug Fix
- [ ] New Feature
- [ ] Performance
- [ ] Refactor (no behaviour change)
- [ ] Documentation
- [ ] Build / CI / dependencies

## How was this tested?

- **Server build tested:** 1a61dc7684792933b2502b7da713f41d6f007901
- **Bedrock client version:** 26.40
- **Plugins loaded:** none

**Steps performed and results:**

1.
2.

- [ ] I built the server and ran it with this change
- [ ] I reproduced the original problem and confirmed this fixes it (bug fixes)
- [ ] Existing unit tests pass (`gradlew test`)
- [ ] I added tests for the new logic, or explained below why that isn't practical

## Breaking changes / API impact

None

## Performance notes

N/A

## AI usage disclosure

| Model | What it did |
|-------|-------------|
|   codex    |   packet analysis       |

- [X] The disclosure above covers **all** AI use in this PR - code, tests, Javadoc, commit messages, config, and research
- [X] I have read every line of this diff myself and can explain any of it
- [X] This PR description and any review replies are written by me, not generated

## Checklist

- [X] My change follows the existing code style
- [X] The PR is one logical change, with no unrelated reformatting or refactors
- [ ] Public API additions/changes have Javadoc
- [X] No dead code, commented-out blocks, or debug logging left behind
- [X] Commit messages follow Conventional Commits
- [X] My contribution is licensed under LGPL-3.0, and any third-party code is attributed
- [X] "Allow maintainer edits" is enabled

PR #3004, opened 2026-08-14
FIX : Reject nameless voxel shapes that broke VoxelShapesPacket encoding

JblusItsMe:
## What does this PR do?

Fixes a crash that prevents VoxelShapesPacket from ever reaching the client.

## Why?

It affects every server on this gamedata version. No plugin is involved and no configuration triggers it — a stock server logs a full stack trace for every single player who joins, and the client silently never gets the shape table it was supposed to receive. Server and client can then disagree on collision and selection boxes for the blocks that table describes.

Bad data should not be able to break a packet. The registry is keyed by String and is handed identifiers straight from a JSON file, so a missing field is a case worth expecting. Object2ObjectOpenHashMap accepting a null key means the problem stays invisible until it surfaces far away, in the serializer, as an NPE with no indication of where the null came from. Validating at register puts the failure where the mistake is, and protects plugins registering their own shapes too.

A single malformed entry should not truncate the registry. The try around the whole loop meant one duplicate key cost 66 perfectly valid shapes, with nothing in the log tying the two together. That failure mode is worse than the crash it caused, because it is silent.

- [X] Bug Fix
- [ ] New Feature
- [ ] Performance
- [ ] Documentation

## How was this tested?

- **Server build tested:** git-5756bf2
- **Bedrock client version:** 1.26.40
- **Plugins loaded:** none

1.
2.

None

## Performance notes

N/A

## AI usage disclosure

| Model | What it did |
|-------|-------------|
|       |             |

## Checklist

- [ ] Public API additions/changes have Javadoc
- [[X] No dead code, commented-out blocks, or debug logging left behind

### DIFF
--- a/src/main/java/org/powernukkitx/registry/VoxelShapeRegistry.java
+++ b/src/main/java/org/powernukkitx/registry/VoxelShapeRegistry.java
@@ -69,16 +69,21 @@ public final class VoxelShapeRegistry implements IRegistry<String, VoxelShapes.S
             final List<VanillaShape> shapes = new Gson().fromJson(reader, new TypeToken<List<VanillaShape>>() {
             }.getType());
             for (VanillaShape shape : shapes) {
+                if (shape.identifier == null || shape.identifier.isBlank()) {
+                    continue;
+                }
                 final List<VoxelBox> boxes = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


sircurse:
It is being addressed already on #3003.

PR #3016, opened 2026-08-15
fix: Always send skin geometry data as a JSON Object

JblusItsMe:
## What does this PR do?

The Bedrock client expects a skin's geometry data to be a JSON object. PR #2977 only covered two cases — null and the empty string — both replaced with {}. Everything else malformed went through untouched: a whitespace-only string, a bare geometry name, a JSON array, or arbitrary bytes base64-decoded from the client in ClientSkinData#readSkin. So the server could still hand the client a value it cannot parse, on both packets that carry a skin (PlayerListPacket and PlayerSkinPacket).

Second, independent reason: EntityHuman#initHumanEntity called builder.skinResourcePatch(GEOMETRY_HUMANOID) instead of builder.geometryData(...) when an entity had no geometry. The 4.6 KB of skin_geometry.json overwrote a valid resource patch, SkinUtils#isValidResourcePatch then failed, and EntityHuman#spawnTo threw IllegalStateException — that entity never appeared for the player.

So: malformed skin data reaching the client, and human entities without geometry data broken outright.

## Why?

Two defects let malformed skin geometry data reach the Bedrock client, which expects that field to hold a JSON object.

#2977 made SkinConverter#toSerializedSkin substitute {} for null and empty geometry data, but only those two cases. Any other malformed value was still forwarded verbatim — a whitespace-only string, a bare geometry name, a JSON array, or arbitrary bytes base64-decoded from the client in ClientSkinData#readSkin. Since toSerializedSkin is the single point where a Skin becomes the SerializedSkin carried by PlayerListPacket and PlayerSkinPacket, every skin sent to a client passes through that gap.

Separately, EntityHuman#initHumanEntity passed GEOMETRY_HUMANOID — the 4.6 KB contents of gamedata/skin_geometry.json — to builder.skinResourcePatch() instead of builder.geometryData() when an entity had no geometry data. That overwrote a valid resource patch with geometry JSON, so SkinUtils#isValidResourcePatch rejected the skin and EntityHuman#spawnTo threw IllegalStateException, leaving the entity invisible to players.

Closes #

## Type of change

- [x] Bug Fix
- [ ] New Feature
- [ ] Performance
- [ ] Documentation

## How was this tested?

- **Server build tested:** git-fbe949f
- **Bedrock client version:** 1.26.44
- **Plugins loaded:** none

1.
2.

No signature changes, no config change, no save-format change. Two behaviour
changes on public methods:

- `SkinConverter#toSerializedSkin` - geometry data that is non-null, non-empty
  but not a JSON object is now sent as `{}` instead of being forwarded verbatim.
  Geometry data that already opens a JSON object is unchanged, apart from
  surrounding whitespace being trimmed.
- `EntityHuman#initHumanEntity` - an entity with no geometry data now keeps its
  resource patch and receives `GEOMETRY_HUMANOID` in `geometryData`. Entity NBT
  written by earlier builds holds the corrupted `SkinResourcePatch`; it is
  repaired on load by the existing `SkinUtils#isValidResourcePatch` branch,
  which substitutes `GEOMETRY_CUSTOM_SLIM`.

## Performance notes

`toSerializedSkin` runs once per player-list entry and per skin broadcast, so it
is on a send path. The change adds one `String#trim()` and one `startsWith` per
call. `trim()` returns the same instance when there is nothing to strip, so the
common case - geometry data that already opens with `{` - allocates nothing. No
benchmark was run; the change is not expected to be measurable.

## AI usage disclosure

| Model | What it did |
|-------|-------------|
|       |             |

## Checklist

- [ ] My contribution is licensed under LGPL-3.0, and any third-party code is attributed

### DIFF
--- a/src/main/java/org/powernukkitx/entity/EntityHuman.java
+++ b/src/main/java/org/powernukkitx/entity/EntityHuman.java
@@ -168,7 +168,7 @@ public class EntityHuman extends EntityHumanType {
             }
 
             if(skin.getGeometryData() == null || skin.getGeometryData().isEmpty()) {
-                builder.skinResourcePatch(GEOMETRY_HUMANOID);
+                builder.geometryData(GEOMETRY_HUMANOID);
                 changed = true;
             }
 
--- a/src/main/java/org/powernukkitx/utils/SkinConverter.java
+++ b/src/main/java/org/powernukkitx/utils/SkinConverter.java
@@ -29,6 +29,7 @@ import java.util.UUID;
 public class SkinConverter {
 
     private final String PERSONA_PREFIX = "persona_";
+    private final String EMPTY_GEOMETRY_DATA = "{}";
     private final int TINT_COLOR_COUNT = 4;
 
     public SerializedSkin toSerializedSkin(Skin skin, boolean trusted) {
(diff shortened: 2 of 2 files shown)


AzaleeX (approved):

xRookieFight (approved):
Please mention the AI usage in your PR description. Next time your PR will be closed.

PleaseInsertNameHere (approved):

PR #3018, opened 2026-08-15
Fix: catch LevelException in block entity update queue [need test]

### LINKED ISSUE #3015: [Bug]: BlockEntityLectern.onUpdate() throws LevelException (Undefined Level reference) during world startup tick
### Describe the bug

During server startup, one of our worlds ("tutorial") fails its initial tick check with an uncaught `LevelException: Undefined Level reference` thrown from a `BlockEntityLectern`'s `isBlockEntityValid()` check. The exception originates from `Position.getValidLevel()` finding `this.level == null` on the block entity's own position, while the block entity is still being processed inside `Level.doTick()`'s pending-block-entity-update queue (`ConcurrentLinkedQueue#removeIf`).

This suggests the lectern's `level` field was cleared (e.g. by a chunk/level unload path) while the block entity remained registered in the level's update queue, so a later `onUpdate()` call runs `isBlockEntityValid()` → `getBlock()` → `getLevelBlock()` → `getValidLevel()` on an already-detached position.

### To Reproduce

We don't have exact reproduction steps (this was caught from a production server's startup log, not a controlled test). It reproducibly happens for a specific `tutorial` world containing at least one lectern, on server startup, every time the server boots.

### Expected behavior

The block entity update queue should not retain (or should defensively skip) block entities whose level reference has already gone away, so `onUpdate()`/`isBlockEntityValid()` never throws for a detached block entity.

### Screenshots

_No response_

### Version

pinned-20260814-c71abad (self-built from a recent source checkout, matches current `main` behavior we observed)

### Additional context

Full log output:

```
2026-08-15 12:44:23 [main] ERROR - ワールド "/home/living/pnx/living/worlds/tutorial" をチェックすることができませんでした: org.powernukkitx.utils.LevelException: Undefined Level reference
    at org.powernukkitx.level.Position.getValidLevel(Position.java:185)
    at org.powernukkitx.level.Position.getLevelBlock(Position.java:143)
    at org.powernukkitx.level.Position.getLevelBlock(Position.java:139)
    at org.powernukkitx.blockentity.BlockEntity.getBlock(BlockEntity.java:180)
    at org.powernukkitx.blockentity.BlockEntityLectern.isBlockEntityValid(BlockEntityLectern.java:60)
    at org.powernukkitx.blockentity.BlockEntity.onUpdate(BlockEntity.java:186)
    at org.powernukkitx.blockentity.BlockEntityLectern.onUpdate(BlockEntityLectern.java:72)
    at org.powernukkitx.level.Level.lambda$doTick$9(Level.java:1503)
    at java.base/java.util.concurrent.ConcurrentLinkedQueue.bulkRemove(ConcurrentLinkedQueue.java:1004)
    at java.base/java.util.concurrent.ConcurrentLinkedQueue.removeIf(ConcurrentLinkedQueue.java:964)
    at org.powernukkitx.level.Level.doTick(Level.java:1503)
    at org.powernukkitx.Server.checkTickUpdates(Server.java:1089)
    at org.powernukkitx.Server.tick(Server.java:1180)
    at org.powernukkitx.Server.tickProcessor(Server.java:1053)
    at org.powernukkitx.Server.start(Server.java:1036)
    at org.powernukkitx.Server.<init>(Server.java:684)
    at org.powernukkitx.PowerNukkitX.main(PowerNukkitX.java:233)
    at org.powernukkitx.JarStart.main(JarStart.java:20)
```

We searched our own plugin codebase (a ~90-plugin Maven multi-module set) and found no plugin code anywhere in the stack or any plugin that touches `BlockEntityLectern`/lecterns directly, so we don't believe this is caused by third-party plugin code — it appears to be a core block-entity lifecycle issue.

iYozemMc:
## What does this PR do?

Wraps the block entity update loop in Level.doTick() in a try/catch so a LevelException from one detached block entity no longer crashes the whole tick.

## Why?

Closes #3015

## Type of change

- [X] Bug Fix
- [ ] New Feature
- [ ] Performance
- [ ] Documentation

## How was this tested?

- **Server build tested:** cdf419a81
- **Bedrock client version:** 26.42
- **Plugins loaded:** NONE

1. need to test

None

## Performance notes

N/A

## AI usage disclosure

| Model | What it did |
|-------|-------------|
|       |             |

## Checklist

### DIFF
--- a/src/main/java/org/powernukkitx/level/Level.java
+++ b/src/main/java/org/powernukkitx/level/Level.java
@@ -1500,7 +1500,13 @@ public class Level implements Metadatable {
                 this.hasAsyncPrepareEntities = seenAsyncPrepare;
             }
             if (prof) phase[5] = -phaseStart + (phaseStart = System.nanoTime());
-            this.updateBlockEntities.removeIf(blockEntity -> !(!blockEntity.closed && blockEntity.isValid() && blockEntity.onUpdate()));
+            this.updateBlockEntities.removeIf(blockEntity -> {
+                try {
+                    return !(!blockEntity.closed && blockEntity.isValid() && blockEntity.onUpdate());
+                } catch (LevelException e) {
+                    return true;
+                }
+            });
             if (prof) phase[6] = -phaseStart + (phaseStart = System.nanoTime());
 
             this.tickChunks();


Byakkono:
Reviewed the diff (not build-tested, since [main] does not build-test PRs by default — happy to actually build/run this against the `tutorial` world repro from #3015 if that'd help).

The fix itself is a safe, minimal stopgap: wrapping the `removeIf` predicate in `try/catch (LevelException)` stops one detached block entity from crashing the entire `doTick()`, which matches the "Expected behavior" section of #3015. It's also scoped to the generic block-entity update loop rather than being lectern-specific, so it protects every block entity type against this failure mode, not just `BlockEntityLectern`.

Two things worth considering before merge:

1. **Cleanup on catch.** The catch branch returns `true` (removes the entity from the queue) but never calls `blockEntity.close()`. `close()` is idempotent (`if (!this.closed) { ... }`), so calling it here would guarantee the entity is also removed from `chunk`/`level.blockEntities` bookkeeping and marked `closed`, instead of only being dropped from `updateBlockEntities`. Right now `close()` is the only place in `BlockEntity`/`Position` that nulls `level`, so this may be moot today — but it'd be a cheap defensive improvement.

2. **Root cause is still open, and there is a real concurrency path for it.** `BlockEntity.level` (inherited from `Position.level`) is a plain, non-`volatile` field with no synchronization. `Level.subTick()` schedules `doLevelGarbageCollection()` on `getServer().getComputeThreadPool()` every 5 ticks — a separate thread from the main tick loop that calls `doTick()`. That's an existing concurrent-access path onto the same block-entity state this update queue iterates over. This PR correctly stops the crash, but doesn't fix why `level` gets nulled while still enqueued — it just contains it. Might be worth logging in the catch (e.g. `log.debug`) so a recurrence is still diagnosable instead of silently disappearing, and/or opening a follow-up issue to track the actual race.

Given the PR checklist is still unchecked and the `tutorial`-world repro from #3015 hasn't been re-tested against this branch yet, I'd hold off merging until that's confirmed — but the change looks safe to land as a stopgap once that's done.


---

**AI disclosure:** Claude Sonnet 5 read the diff and the linked source files (`BlockEntity.java`, `BlockEntityLectern.java`, `Position.java`, `Level.java`) and drafted the analysis above; it was not build-tested. Missed disclosing this when I first posted — adding it now per CONTRIBUTING.md.

PR #3058, opened 2026-08-20
fix: wait for a chunk to be generated before populating what a structure put there

Wyze3306:
## What does this PR do?

In the title

## Why?

Create a bug with infinite entity spawn in the void.

No linked issue.

## Type of change

- [x] Bug Fix
- [ ] New Feature
- [ ] Performance
- [ ] Documentation

## How was this tested?

- **Server build tested:** `2a7f4d4b2` (branch `fix/structure-hooks-ungenerated-chunk`, one commit on top of upstream `beta` `7a93f60a0`)
- **Bedrock client version:** none yet, automated tests only

Automated only, so far:
- `./gradlew compileJava compileTestJava` passes
- `./gradlew test` on `2a7f4d4b2`: **1066 tests, 0 failures, 0 errors**
- `BlockManagerChunkHookTest` covers a hook on a generated chunk, a hook waiting for its chunk, a waiting hook running after the blocks held with it, hooks merged from another manager, a waiting hook that throws, clearing the hooks of a level, and the bound on waiting chunks

Adds to `BlockManager`: `addHook(int, int, Runnable)`, `addHook(Block, Runnable)`,
`addHook(BlockVector3, Runnable)`, `mergeHooks(BlockManager)` and the static `clearPendingHooks(int)`.
The existing `addHook(Runnable)` keeps its immediate behaviour. Adds `byChunk()` to
`EndCityPieces.PostPlacement` and `WoodlandMansionPieces.PostPlacement`. The protected
`VillageStructure.populatePendingChestLoot(Level)` becomes `populateChestLoot(Level, BlockVector3, RandomizableContainer)`,
which fills one chest. No config or save-format change.

Known limit: a hook waiting on a chunk only lives in memory, whereas the blocks it waits with are saved with
the chunk. If the server stops, or the level is unloaded, before that chunk is generated, the blocks are
still placed once it is, but the hook never runs.

## Performance notes

N/A

## AI usage disclosure

| Model | What it did |
|-------|-------------|
| Claude Opus 5 (Claude Code, autonomous agent) | Rebased my fork on `upstream/master`, split my pre-existing fixes into one branch per bug, removed my explanatory comments and Javadoc from the diff, translated my French commit messages into English, and ran the build and the test suite. |

## Checklist


Buddelbubi:
When would that apply? BlockManager already ensures that every chunk where blocks are placed is generated before. And I dont think any Structure spawns in entities where it did not place a structure.

Wyze3306:
For example, I //cut a large area, and this problem occurred.

Buddelbubi:
It uses a hook? Why? And why in places where no blocks were changed? 
I feel like thats a plugin bug

src/main/java/org/powernukkitx/level/generator/object/BlockManager.java
```diff
@@ -19,26 +19,46 @@
...
     private static final String PENDING_SUB_CHUNK_UPDATES = "pendingSubChunkUpdates";
 
+    private static final int MAX_PENDING_HOOK_CHUNKS = 4096;
+
+    private static final LinkedHashMap<PendingHookKey, List<Runnable>> PENDING_HOOKS = new LinkedHashMap<>() {
+        @Override
+        protected boolean removeEldestEntry(Map.Entry<PendingHookKey, List<Runnable>> eldest) {
+            return size() > MAX_PENDING_HOOK_CHUNKS;
+        }
+    };
```
xRookieFight, line 59: `removeEldestEntry` silently discards the oldest chunk's hooks once 4096 chunks are pending. Those are chest fills, entity spawns, and block updates: a busy server crossing the cap loses structure contents with no log line. Either log the eviction, or bound it per level instead of dropping work the generator committed to.

src/main/java/org/powernukkitx/level/generator/object/BlockManager.java
```diff
@@ -51,21 +71,102 @@ public BlockManager(Level level) {
...
+    }
+
+    public static LongOpenHashSet chunkHashesOfPositions(Collection<BlockVector3> positions) {
+        LongOpenHashSet hashes = new LongOpenHashSet();
+        for (BlockVector3 pos : positions) {
+            hashes.add(Level.chunkHash(pos.getX() >> 4, pos.getZ() >> 4));
+        }
+        return hashes;
+    }
+
```
xRookieFight, line 170: this.level.getChunk(x, z) falls through to forceLoadChunk (Level.java:4942-4946), so the "is it generated" probe does a synchronous disk load on the generator thread for every unloaded chunk hook. getChunkIfLoaded gives the same answer with no I/O - a null there means "not ready", which is exactly the defer case.

src/main/java/org/powernukkitx/level/generator/object/BlockManager.java
```diff
@@ -19,26 +19,46 @@
...
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Predicate;
 
 public class BlockManager {
     private static final String PENDING_SUB_CHUNK_UPDATES = "pendingSubChunkUpdates";
 
+    private static final int MAX_PENDING_HOOK_CHUNKS = 4096;
+
+    private static final LinkedHashMap<PendingHookKey, List<Runnable>> PENDING_HOOKS = new LinkedHashMap<>() {
```
xRookieFight, line 46: PENDING_HOOKS is static and only ever drained by runPendingHooks on generation. Hooks for chunks a player never visits stay forever, and nothing purges on level unload, so entries survive their levelId and can fire against a different world if the id is reused. Needs a clear on level unload.

src/main/java/org/powernukkitx/level/generator/object/BlockManager.java
```diff
@@ -51,21 +71,102 @@ public BlockManager(Level level) {
...
+    }
+
     public ObjectOpenHashSet<Runnable> getHooks() {
         return this.hooks;
     }
 
+    public Long2ObjectOpenHashMap<ObjectArrayList<Runnable>> getChunkHooks() {
+        return this.chunkHooks;
+    }
+
```
xRookieFight: getChunkHooks() has no caller. Drop it rather than shipping unused public API.

src/main/java/org/powernukkitx/level/generator/object/BlockManager.java
```diff
@@ -51,21 +71,102 @@ public BlockManager(Level level) {
...
+    public void addHook(LongCollection chunkHashes, Runnable runnable) {
+        LongOpenHashSet distinct = new LongOpenHashSet(chunkHashes);
+        if (distinct.isEmpty()) {
+            this.addHook(runnable);
+            return;
+        }
+        AtomicInteger remaining = new AtomicInteger(distinct.size());
+        for (long chunkHash : distinct) {
+            this.chunkHooks.computeIfAbsent(chunkHash, k -> new ObjectArrayList<>()).add(() -> {
+                if (remaining.decrementAndGet() == 0) {
```
xRookieFight, line 150: addHook(LongCollection, Runnable): if any one chunk in the set never generates, remaining never reaches 0 and the runnable never runs, while the other chunks' entries are consumed and removed. Worth a comment stating that all-or-nothing is intended, since the failure is invisible.

xRookieFight (changes requested):
Also, all seven new public methods plus the two statics in BlockManager ship without Javadoc. The PR checklist calls this out as not done; the repo requires it for public API.

xRookieFight (commented):
Diagnosis looks right to me, and the `PopulatorStage` change is a real fix since `applySubChunkUpdate()` already calls `setChanged()` itself. The `PillagerOutpostPopulator` one stands on its own too: the old `scheduleDelayedTask(..., 10)` registered its hook long after `applyHooks()` had run, so those golems and allays never spawned at all. Three notes:

**`BlockManager.java:196` and `:212`**

Check then defer isn't atomic. `applyHooks()` reads `chunk.isGenerated()` and only afterwards calls `deferHooks()`, which takes the `PENDING_HOOKS` lock. Another generation thread can finish that same chunk in between: `GeneratedStage` flips the state, `runPendingHooks()` takes the lock, finds nothing, returns. The deferring thread then inserts an entry nobody comes back for, and it sits until the 4096 eviction drops it. Two chunks generating at once is the normal case here. Re-checking `isGenerated()` inside the `synchronized (PENDING_HOOKS)` block and running inline if it flipped would close it.

**`BlockManager.java:44`**

Deferred hooks don't survive a restart but the blocks they belong to do. Blocks wait in the chunk's `pendingSubChunkUpdates` NBT, which is saved; `PENDING_HOOKS` is a static in memory map. So if the server stops before the neighbour chunk generates, the chunk comes back with the structure's blocks and no hook: chests with no loot, banners with no type, no outpost golem. That's arguably worse than today's bug, which at least leaves the structure visibly unfinished instead of looking complete and being empty. Worth deciding whether the hooks need to ride along with the blocks or whether this gets documented as a known limit.

**`BlockManager.java:141`**

All or nothing plus eviction means one lost chunk kills the whole structure. `addHook(LongCollection, ...)` only fires on the last chunk, so if any one of them is evicted at `MAX_PENDING_HOOK_CHUNKS` the counter never reaches zero and the hooks that already fired for the other chunks did nothing. The warning at line 48 names one chunk, which makes it read like one chunk was affected when the real blast radius is every chunk that structure covered.

PR #3064, opened 2026-08-20
fix: removed unused setting allowNether and allowTheEnd

Buddelbubi:
removes unused settings allowNether and allowTheEnd.
Those dimensions are disabled by not adding them to the worlds config.json


xRookieFight (changes requested):
Why don't we use it? Currently it creates a new end and nether to every level default. Atleast we should only create The End and Nether in normal vanilla generator. The others should be the user's choice

Buddelbubi:
Its vanilla. Thats why. 
Those settings are unused. Implementing them would be worse since its two ways in doing the same thing which could contradict themselves.

Vanilla generates three dimensions for a world. So we do it too. If you do not want end or nether, just remove the dimensions from the worlds config.json. 
It already is the users choise which dimensions to use. Please remove your change request. @xRookieFight

xRookieFight (approved):

PleaseInsertNameHere (approved):

PR #3081, opened 2026-08-23
fix: Patch custom craft with custom item result

UnknownGuyEXE:
## What does this PR do?

my pr fixes custom items in craft results

## Why?

bugfix

## Type of change

- [X] Bug Fix
- [ ] New Feature
- [ ] Performance
- [ ] Documentation

## How was this tested?

- **Server build tested:** 35d0be78d
- **Bedrock client version:** 26.44
- **Plugins loaded:** custom plugin

1. join a server with a recipe that gives a custom item as the result
2. check if the craft is shown in the preselection

None

## Performance notes

N/A

## AI usage disclosure

no ai used

## Checklist

### DIFF
--- a/src/main/java/org/powernukkitx/registry/CreativeItemRegistry.java
+++ b/src/main/java/org/powernukkitx/registry/CreativeItemRegistry.java
@@ -9,6 +9,7 @@ import it.unimi.dsi.fastutil.objects.ObjectLinkedOpenHashSet;
 import lombok.extern.slf4j.Slf4j;
 import org.cloudburstmc.nbt.NbtMap;
 import org.cloudburstmc.nbt.NbtUtils;
+import org.cloudburstmc.protocol.bedrock.data.inventory.ItemData;
 import org.cloudburstmc.protocol.bedrock.data.payload.creative.CreativeGroupInfoPayload;
 import org.cloudburstmc.protocol.bedrock.data.payload.creative.CreativeItemCategory;
 import org.cloudburstmc.protocol.bedrock.data.payload.creative.CreativeItemEntryPayload;
@@ -311,9 +312,20 @@ public class CreativeItemRegistry implements ItemID, IRegistry<Integer, Item, It
         }
         final CreativeItemEntryPayload entryPayload = new CreativeItemEntryPayload();
         entryPayload.setCreativeNetId(new CreativeItemNetId(MAP.isEmpty() ? 0 : MAP.lastIntKey()));
-        entryPayload.setItemInstance(value.toNetwork());
+        entryPayload.setItemInstance(toCreativeItemData(value));
         entryPayload.setGroupIndex(groupIndex);
         ITEM_DATA.add(entryPayload);
+    }
+
+    private static ItemData toCreativeItemData(Item value) {
+        final CompoundTag nbt = value.getNbt();
+        if (nbt != null && nbt.contains("Damage") && nbt.getInt("Damage") == 0) {
+            final Item pristine = value.clone();
+            final CompoundTag stripped = nbt.copy().remove("Damage");
+            pristine.setNbt(stripped.isEmpty() ? null : stripped);
+            return pristine.toNetwork();
+        }
+        return value.toNetwork();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/org/powernukkitx/registry/CreativeItemRegistry.java
```diff
@@ -311,11 +312,22 @@ public void register(Integer key, Item value, int groupIndex) throws RegisterExc
...
+        final CompoundTag nbt = value.getNbt();
+        if (nbt != null && nbt.contains("Damage") && nbt.getInt("Damage") == 0) {
+            final Item pristine = value.clone();
+            final CompoundTag stripped = nbt.copy().remove("Damage");
+            pristine.setNbt(stripped.isEmpty() ? null : stripped);
+            return pristine.toNetwork();
+        }
+        return value.toNetwork();
+    }
+
```
xRookieFight: `toCreativeItemData` is a verbatim copy of the Damage-stripping block already in Item.toRecipeNetwork() (Item.java:2719-2727), down to the isEmpty() ? null and the nbt.copy().remove("Damage"). Extract it once on Item (e.g. a private strippedForNetwork()) and call it from both.

src/main/java/org/powernukkitx/registry/CreativeItemRegistry.java
```diff
@@ -311,11 +312,22 @@ public void register(Integer key, Item value, int groupIndex) throws RegisterExc
...
         entryPayload.setGroupIndex(groupIndex);
         ITEM_DATA.add(entryPayload);
     }
 
+    private static ItemData toCreativeItemData(Item value) {
+        final CompoundTag nbt = value.getNbt();
+        if (nbt != null && nbt.contains("Damage") && nbt.getInt("Damage") == 0) {
+            final Item pristine = value.clone();
+            final CompoundTag stripped = nbt.copy().remove("Damage");
+            pristine.setNbt(stripped.isEmpty() ? null : stripped);
```
xRookieFight: `pristine.toNetwork()` on a clone. toNetwork() auto-assigns a stack net id (Item.java:2699-2701), so the id is burned on the throwaway clone and the item actually stored in MAP never gets one, unlike every non-Damage item that goes through the value.toNetwork() branch. Assign on value, or use a path that does not allocate a net id.

src/main/java/org/powernukkitx/registry/CreativeItemRegistry.java
```diff
@@ -311,11 +312,22 @@ public void register(Integer key, Item value, int groupIndex) throws RegisterExc
         }
         final CreativeItemEntryPayload entryPayload = new CreativeItemEntryPayload();
         entryPayload.setCreativeNetId(new CreativeItemNetId(MAP.isEmpty() ? 0 : MAP.lastIntKey()));
-        entryPayload.setItemInstance(value.toNetwork());
+        entryPayload.setItemInstance(toCreativeItemData(value));
```
xRookieFight: Item.toCreativeNetwork() (Item.java:2744) exists for exactly this payload and already sets usingNetId(false) / netId(0). The registry calls toNetwork() instead. If the Damage strip were added there, both call sites would just be value.toCreativeNetwork() and the net-id issue above disappears. Worth checking with the author whether toCreativeNetwork was meant to be used here.

src/main/java/org/powernukkitx/registry/CreativeItemRegistry.java
```diff
@@ -325,7 +337,7 @@ public void register(Integer key, Item value) throws RegisterException {
         } else {
             final CreativeItemEntryPayload entryPayload = new CreativeItemEntryPayload();
             entryPayload.setCreativeNetId(new CreativeItemNetId(MAP.isEmpty() ? 0 : MAP.lastIntKey()));
-            entryPayload.setItemInstance(value.toNetwork());
+            entryPayload.setItemInstance(toCreativeItemData(value));
```
xRookieFight: Item.toCreativeNetwork() (Item.java:2744) exists for exactly this payload and already sets usingNetId(false) / netId(0). The registry calls toNetwork() instead. If the Damage strip were added there, both call sites would just be value.toCreativeNetwork() and the net-id issue above disappears. Worth checking with the author whether toCreativeNetwork was meant to be used here.

xRookieFight (dismissed):

sircurse (approved):

PleaseInsertNameHere (approved):

Buddelbubi (approved):

xRookieFight (approved):

PR #3087, opened 2026-08-25
fix: stop queueing packets for a player whose session is gone

Wyze3306:
A closed session never drains its send queue - BedrockPeer.onTick returns on the closed flag - and that queue is unbounded. Every packet still handed to a player whose connection is already gone therefore stays in memory until the server stops, and one player left behind in a busy world collects every broadcast aimed at it: three of them held 5.7 million packets between them. sendPacket now drops a packet aimed at a session that is no longer connected, as sendPacketImmediately already does. The test is on the session alone rather than on isConnected(), because close() clears the connected flag on entry and still has packets to send while it tears the player down.

Two paths left such a player behind. The network thread can only ask for a close by setting pendingClose, and onUpdate read it from inside the branch that handles a spawned, living player. A player who lost the connection while dead took the branch above it, which always returns, so the close never happened: the player stayed in Level.players with a dead peer for the rest of the server's life. The check now runs before anything that can return early.

PlayerSessionHolder.doPlayerCreation went on to processLogin even when Server.onPlayerLogin had cancelled the login, closed the player and taken it out of the server lists. processLogin ends in Entity.init, which puts the player back into Level.players, and nothing takes it out again because close() cannot run a second time.

It's a bug of RAM.

### DIFF
--- a/src/main/java/org/powernukkitx/Player.java
+++ b/src/main/java/org/powernukkitx/Player.java
@@ -1194,6 +1194,16 @@ public class Player extends EntityHuman implements CommandSender, ChunkLoader, I
         this.pendingClose = reason;
     }
 
+    private boolean closeIfRequested() {
+        final String closeReason = this.pendingClose;
+        if (closeReason == null) {
+            return false;
+        }
+        this.pendingClose = null;
+        this.close(closeReason);
+        return true;
+    }
+
     /**
      * Offers a new movement task to the player, considering distance and rotation thresholds.
      * Also handles the special case where an erroneous position may be received right after teleportation.
@@ -2992,6 +3002,11 @@ public class Player extends EntityHuman implements CommandSender, ChunkLoader, I
      * @param packet packet to send
      */
     public void sendPacket(BedrockPacket packet) {
+        // Deliberately the session's flag and not isConnected(): close() clears the player's own
+        // flag on entry and still sends packets while it tears the player down.
+        if (!this.session.isConnected()) {
+            return;
+        }
         final PacketSendEvent event = new PacketSendEvent(this, packet);
         this.server.getPluginManager().callEvent(event);
         if (event.isCancelled()) {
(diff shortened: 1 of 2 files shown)


AzaleeX (approved):

PleaseInsertNameHere (approved):

src/main/java/org/powernukkitx/Player.java
```diff
@@ -3444,6 +3447,13 @@ public boolean onUpdate(int currentTick) {
             }
         }
 
+        if (this.pendingClose != null) {
```
xRookieFight: Close runs before drainInboundPackets(), contradicting the requestClose javadoc at Player.java:1187-1190. Move the pendingClose block below the drain, or update that javadoc.

src/main/java/org/powernukkitx/Player.java
```diff
@@ -2992,6 +2992,9 @@ public void updateTrackingPositions(boolean delayed) {
      * @param packet packet to send
      */
     public void sendPacket(BedrockPacket packet) {
+        if (!this.session.isConnected()) {
+            return;
+        }
```
xRookieFight, line 3009: Please add a comment saying the check is on session.isConnected() and not isConnected() on purpose, since close() sends packets with connected already false.

xRookieFight (approved):
Some nits. LGTM

xRookieFight (commented):
Two things:

**PlayerSessionHolder.java:460** - the guard skips `processLogin()`, but doesn't undo what `onPlayerLogin` already did. On the non-cancelled path `Server.java:1706` runs `players.put(socketAddress, player)` regardless, so if a `PlayerLoginEvent` handler kicks the player itself instead of cancelling, `close()` runs first (its `removePlayer` finds nothing, the player isn't in the map yet) and then `put()` inserts the closed player. Nothing removes it after that, since `close()` won't run twice. Same leak, one map up. Mirroring the cancel-path cleanup before returning would cover it.

**PR description** - it says `sendPacket` now drops "as sendPacketImmediately already does", but that one checks `isConnected()` (player flag + session) while the new code checks only `session.isConnected()`. The difference is on purpose and the inline comment says why - it's just the description that reads like they match. Worth correcting so nobody later harmonises the two and breaks sends during `close()`.

Buddelbubi (approved):

Buddelbubi:
Is there a reason why its not merged yet?

PR #3088, opened 2026-08-25
security: inbound packet validation

Wyze3306:
## What does this PR do?

Fixes six bugs in the inbound packet handlers, all reachable by an ordinary client with no operator
rights and no plugin involved: indices that go straight from the packet into an inventory, reach
checks that were far too loose or ran after the work they were meant to guard, and a duplication
through the block pick.

## Why?

I went through the handlers that take a position, a slot or an index from the client and use it
without bounding it first. Six of them do, and they fail in different ways.

**Duplication and permission**

- `BlockPickRequestPacket` carries a client controlled `withData` flag that was honoured for every
  game mode and every block entity. `getCleanedNBT()` only strips `id`/`x`/`y`/`z`, so the whole
  tile NBT ended up in the item's `BlockEntityTag`, and the `place()` of most blocks copies that tag
  back into the new tile. Picking a filled chest, barrel, shulker box, hopper, furnace or brewing
  stand therefore recreated the container with its contents, and item frames, chiseled bookshelves,
  campfires and beehives leaked their items, food or bees the same way. What a pick copies is now
  decided by the block entity, through `getPickNBT(Player)`, and nothing is copied unless it opts in.
- `BlockCommandBlock.place()` only requires `isCreative()`, not `isOp()`, so a creative non
  operator could pick an operator's command block with data and place copies that still carry its
  command, bypassing the `isOp() && isCreative()` gate `CommandBlockUpdateHandler` enforces to
  write one. Command blocks, and structure blocks which vanilla restricts to operators the same
  way, now only copy their data for an operator.

**Indices that reach the wrong slots**

- Both pick handlers scanned to `getInventory().getSize()`, which is 40 and covers the armor slots
  at `ARMORS_INDEX`. A single packet could pull a piece of armor out of its slot, curse of binding
  included, or push an arbitrary item into one, without going through any transaction validation.
- `BookEditPacket` passed its page index and page text straight to `ItemBookWritable`, whose
  `Preconditions` throw `IllegalArgumentException` outside `0..49` or past 256 characters, so a
  malformed packet left the handler through an exception rather than a rejection. Its book slot was
  unbounded too: `setItem()` accepts the armor slots.
- `CrafterInventory.setSlotState()` computes `1 << slot` and its `@Range(from = 0, to = 8)` is not
  enforced at runtime, so a client controlled index corrupted the disabled slot mask, negative and
  out of range values included, and each call broadcasts a spawn packet and updates the surrounding
  redstone.

**Reach and work done before validation**

- `PlayerToggleCrafterSlotRequestPacket` resolved the block with `Level.getBlock(Vector3)`, which
  loads on miss, from unvalidated coordinates: spamming it with far away positions forced
  synchronous chunk loads anywhere in the world, and `getOrCreateBlockEntity()` then created a tile
  there.
- `BlockActorDataPacket` used `distanceSquared > 10000`, about 100 blocks, and `BlockPickRequest`
  used `> 1000`, about 31, both with no angle check.
- `PlayerSkinPacket` converted the serialized skin and wrapped it in a `Skin` before the
  spawned/alive guard and before the cooldown check, so a client could make the server decode a full
  skin payload on every packet even though the update was going to be dropped. The cooldown
  rejection was also logged at `warn`, so spamming the packet flooded the console.
- The pick and NBT paths ran for every game mode even though the item can only be granted in
  creative, so a survival player could force `saveNBT()` plus a deep copy of a large container at
  packet rate.

Closes #

## Type of change

- [x] Bug Fix
- [ ] New Feature
- [ ] Performance
- [ ] Documentation

## How was this tested?

- **Server build tested:**
- **Bedrock client version:**
- **Plugins loaded:**

1.
2.

Automated:
- `./gradlew compileJava compileTestJava` passes on every commit of the branch
- `./gradlew test` on `492ae9136`: **1064 tests, 0 failures, 0 errors**
- `BlockEntityPickNbtTest` checks that a chest, a jukebox and a campfire copy nothing, that a sign copies its data, and that command and structure blocks copy theirs for an operator only

| Change | Impact |
|---|---|
| `Player.canInteract(Vector3)` | New overload carrying the creative/survival block reach, so the `isCreative() ? 13 : 7` that was written out at five call sites is stated once. Addition only, nothing removed or changed in signature. |
| `PlayerSkinPacket` on cooldown | A skin update rejected for being on cooldown no longer fires a pre-cancelled `PlayerChangeSkinEvent`. Building that event needs the decoded `Skin`, which is exactly the work the cooldown check now avoids, so a plugin that reacted to it, or un-cancelled it, no longer sees the rejection. Every other path still fires the event as before. |
| `BlockEntity.getPickNBT(Player)` | New, returns `null` by default: it is what a block pick with data copies onto the item. Signs, skulls, flower pots, cauldrons, enchanting tables, ender chests and mob spawners return their data, command and structure blocks only for an operator, and every other block entity copies nothing. A plugin block entity that wants its data kept on a pick overrides it. |
| `ItemBookWritable.MAX_PAGES`, `ItemBookWritable.MAX_PAGE_LENGTH` | New public constants, checked by `ItemBookWritable` itself and by `BookEditHandler`. |
| `BlockPickRequestPacket`, `ActorPickRequestPacket` | `withData` is honoured in creative only, and the inventory scan stops at `ARMORS_INDEX`, so a pick can no longer move a piece of armor in or out of its slot. |
| `BookEditPacket` | The book slot is bounded to the hotbar, which is the only place a book is edited from. A packet naming a slot past it is dropped instead of writing there. |
| `BlockActorDataPacket`, `BlockPickRequestPacket`, `PlayerToggleCrafterSlotRequestPacket` | Reach is now `canInteract()`, 7 blocks and 13 in creative with the angle check, instead of 31 to 100 blocks with none. A plugin or client acting from further away is refused. |
| All six handlers | They now return early for a player that is not spawned or is dead, as the other handlers already did. |

## Performance notes

Nothing here costs anything, and three paths do less work than before:

- `PlayerSkinHandler` no longer decodes the serialized skin for an update it is going to reject.
- `PlayerToggleCrafterSlotRequestHandler` reads the loaded chunk instead of loading one, so the
  packet can no longer trigger synchronous chunk loads.
- The pick handlers no longer deep copy entity or tile NBT outside creative, where the result was
  discarded anyway.

## AI usage disclosure

| Model | What it did |
|-------|-------------|
|       |             |

## Checklist

- [ ] My contribution is licensed under LGPL-3.0, and any third-party code is attributed
- [ ] "Allow maintainer edits" is enabled


AzaleeX (approved):

PleaseInsertNameHere (changes requested):
Please follow our contributing guidelines

src/main/java/org/powernukkitx/network/process/handler/PlayerSkinHandler.java
```diff
@@ -24,26 +24,28 @@ public class PlayerSkinHandler implements PacketHandler<PlayerSkinPacket> {
...
             log.debug("Player {} tried to update skin while not spawned or dead", playerHandle.getUsername());
             return;
         }
 
+        long cooldown = TimeUnit.SECONDS.toMillis(player.getServer().getSettings().playerSettings().skinChangeCooldown());
+        if (cooldown > System.currentTimeMillis() - player.lastSkinChange) {
+            log.debug("Player {} changed skin too quick!", playerHandle.getUsername());
+            return;
+        }
+
```
xRookieFight, line 38: Cooldown now returns before PlayerChangeSkinEvent is fired. Before, the event still fired (pre-cancelled) and plugins could uncancel or react. Silent drop is an API break. Either keep firing it cancelled, or call it out in the PR description.

src/main/java/org/powernukkitx/network/process/handler/PlayerToggleCrafterSlotRequestHandler.java
```diff
@@ -15,19 +15,34 @@
  */
 public class PlayerToggleCrafterSlotRequestHandler implements PacketHandler<PlayerToggleCrafterSlotRequestPacket> {
 
+    private static final int CRAFTER_SLOT_COUNT = 9;
+
```
xRookieFight: CRAFTER_SLOT_COUNT = 9 duplicates the size in CrafterInventory.java:27. Resolve the crafter first, then bound slot against blockEntity.getInventory().getSize().

src/main/java/org/powernukkitx/network/process/handler/ActorPickRequestHandler.java
```diff
@@ -25,6 +25,10 @@ public class ActorPickRequestHandler implements PacketHandler<ActorPickRequestPa
     public void handle(ActorPickRequestPacket packet, PlayerSessionHolder holder, Server server) {
         final PlayerHandle playerHandle = holder.getPlayerHandle();
         final Player player = playerHandle.player;
+        if (!player.spawned || !player.isAlive()) {
+            log.debug("Player {} tried to send an entity pick request while not spawned or dead", playerHandle.getUsername());
+            return;
+        }
```
xRookieFight, line 31: No distance check added. player.level.getEntity(packet.getActorID()) still resolves any entity ID in the level, so a client can pick an entity across the world. Same class of bug the other three handlers just fixed.

src/main/java/org/powernukkitx/network/process/handler/BlockActorDataHandler.java
```diff
@@ -24,18 +24,20 @@ public void handle(BlockActorDataPacket packet, PlayerSessionHolder holder, Serv
...
-        if (t instanceof BlockEntitySpawnable) {
+        BlockEntity t = player.level.getBlockEntity(position);
+        if (t instanceof BlockEntitySpawnable spawnable) {
+            player.resetInventory();
+
             NbtMap nbt = packet.getActorDataTags();
-            if (!((BlockEntitySpawnable) t).updateCompoundTag(nbt, player)) {
-                ((BlockEntitySpawnable) t).spawnTo(player);
+            if (!spawnable.updateCompoundTag(nbt, player)) {
+                spawnable.spawnTo(player);
```
xRookieFight, line 39: `resetInventory()` moved inside the instanceof BlockEntitySpawnable branch, so a packet naming a missing or non-spawnable block entity no longer resyncs the inventory. If that desync guard was the point, it should stay outside.

src/main/java/org/powernukkitx/network/process/handler/BlockActorDataHandler.java
```diff
@@ -24,18 +24,20 @@ public void handle(BlockActorDataPacket packet, PlayerSessionHolder holder, Serv
             return;
         }
 
-        Vector3 pos = new Vector3(packet.getBlockPosition().getX(), packet.getBlockPosition().getY(), packet.getBlockPosition().getZ());
-        if (pos.distanceSquared(player) > 10000) {
+        BlockVector3 position = BlockVector3.fromNetwork(packet.getBlockPosition());
+        player.temporalVector.setComponents(position.x, position.y, position.z);
+        if (!player.canInteract(player.temporalVector.add(0.5, 0.5, 0.5), player.isCreative() ? 13 : 7)) {
```
xRookieFight: player.isCreative() ? 13 : 7 now appears in five places counting Player.java:659 and InventoryTransactionHandler.java:303. Pull it into a constant or a player.canInteract(pos) overload.

src/main/java/org/powernukkitx/network/process/handler/BlockPickRequestHandler.java
```diff
@@ -26,26 +27,33 @@ public class BlockPickRequestHandler implements PacketHandler<BlockPickRequestPa
...
         if (!player.spawned || !player.isAlive()) {
             log.debug("Player {} tried to send a block pick request while not spawned or dead", playerHandle.getUsername());
             return;
         }
 
-        if (block.distanceSquared(player) > 1000) {
+        BlockVector3 position = BlockVector3.fromNetwork(packet.getPosition());
+        player.temporalVector.setComponents(position.x, position.y, position.z);
+
+        if (!player.canInteract(player.temporalVector.add(0.5, 0.5, 0.5), player.isCreative() ? 13 : 7)) {
```
xRookieFight: player.isCreative() ? 13 : 7 now appears in five places counting Player.java:659 and InventoryTransactionHandler.java:303. Pull it into a constant or a player.canInteract(pos) overload.

src/main/java/org/powernukkitx/network/process/handler/PlayerToggleCrafterSlotRequestHandler.java
```diff
@@ -15,19 +15,34 @@
...
+        int slot = packet.getSlotIndex();
+        if (slot < 0 || slot >= CRAFTER_SLOT_COUNT) {
+            return;
+        }
+
         BlockVector3 position = BlockVector3.fromNetwork(packet.getPos());
-        Block block = level.getBlock(position.asVector3());
+        player.temporalVector.setComponents(position.x, position.y, position.z);
+        if (!player.canInteract(player.temporalVector.add(0.5, 0.5, 0.5), player.isCreative() ? 13 : 7)) {
```
xRookieFight: player.isCreative() ? 13 : 7 now appears in five places counting Player.java:659 and InventoryTransactionHandler.java:303. Pull it into a constant or a player.canInteract(pos) overload.

src/main/java/org/powernukkitx/network/process/handler/BookEditHandler.java
```diff
@@ -20,11 +21,24 @@
...
 
+        if (!player.spawned || !player.isAlive()) {
+            log.debug("Player {} tried to edit a book while not spawned or dead", playerHandle.getUsername());
+            return;
+        }
+
+        if (packet.getBookSlot() < 0 || packet.getBookSlot() >= HumanInventory.ARMORS_INDEX) {
+            log.debug("{}: BookEditPacket with an out of range book slot {}", playerHandle.getUsername(), packet.getBookSlot());
+            return;
+        }
```
xRookieFight: Slot bound is ARMORS_INDEX (36), but book editing only happens from the hotbar. getHotbarSize() is the tighter bound.

xRookieFight (changes requested):

xRookieFight (commented):
Reach checks, the `ARMORS_INDEX` bound and the crafter slot bound all look right to me. Four notes:

**`BlockPickRequestHandler.java:47`**

Two tiles still slip through the denylist. Jukebox keeps its disc under `RecordItem` (`BlockEntityJukebox.java:82`) and campfire keeps food under `Item1`..`Item4` (`BlockEntityCampfire.java:134`), so both still dupe on place. A denylist will need updating every time someone adds a tile, so an allowlist or a per tile `getPickNBT()` hook would age better.

**`BlockActorDataHandler.java:28`, `BlockPickRequestHandler.java:36`, `PlayerToggleCrafterSlotRequestHandler.java:27`**

`temporalVector` isn't buying anything here. `Vector3.add` returns a new instance (`Vector3.java:127`), so you allocate anyway and just leave the temporal vector holding stale values for whoever reads it next. `new Vector3(position.x + 0.5, ...)` is the same cost without touching entity state from a packet handler.

**`BlockPickRequestHandler.java:44`**

`BlockEntityStructBlock` is op only in vanilla for the same reason command blocks are, so it probably wants the same gate.

**`BookEditHandler.java:23`**

`MAX_PAGES` and `MAX_PAGE_LENGTH` restate the literals already sitting in `ItemBookWritable.java:31` and `:64`, and if either side moves the handler starts rejecting valid packets. Worth exposing them from `ItemBookWritable`. Also the `title.length() > 512` check at line 48 can never fire now that `FINALIZE` rejects anything past 64.

PleaseInsertNameHere (approved):

PR #3097, opened 2026-08-25
fix: improve player chunk resynchronization

iYozemMc:
## What does this PR do?

improves player chunk loading and resynchronization by adjusting the initial chunk loading radius retrying the current chunk when needed and periodically resending already loaded chunks when the player is fully synchronized it also improves queue processing so unavailable chunks do not block other chunks from being processed

## Why?

this addresses cases where chunks can remain missing or become out of sync for the player even though the server has finished processing its chunk queues ### ** and I'm not sure because the bug isn't 100% fixed, but it seems to have fixed the bug where it wouldn't reload anymore**

## Type of change

- [X] Bug Fix
- [ ] New Feature
- [X] Performance
- [ ] Documentation

## How was this tested?

- **Server build tested:** ff9b475e7
- **Bedrock client version:** 26.44
- **Plugins loaded:** None

1. i join the server and fly for 10 minutes

None

## Performance notes

Measured on JDK 21 with the project's fastutil (8.5.19). These are **isolated
micro-benchmarks / faithful simulations** of the changed logic, not an end-to-end
server benchmark — each one states what it measures and what it assumes.

### 1. `removeOutOfRadiusChunks` — zero-allocation rewrite
Runs on every chunk-boundary crossing, per player. Allocation measured
deterministically via `ThreadMXBean`.

| View distance | Chunks tracked | Before (alloc/call) | After (alloc/call) |
|---:|---:|---:|---:|
| 8  | 197   | 488 B   | **0 B** |
| 16 | 797   | 768 B   | **0 B** |
| 24 | 1 793 | 1 176 B | **0 B** |
| 32 | 3 209 | 1 736 B | **0 B** |

One `LongArrayList` (+ backing array) per call → **fully eliminated** (single
pass with `iterator.remove()`). Removal CPU time is within noise; the win is GC
pressure, which scales with player count × movement.

### 2. `loadQueuedChunks` — budget no longer wasted on not-ready chunks
Scheduling simulation of a fresh view load, async load latency 0–6 ticks.
"Ticks to fully load the view" (lower = faster):

| VD | Budget | Chunks | Before | After | Improvement |
|---:|---:|---:|---:|---:|---:|
| 8  | 4 | 197   | 192 | **71**  | −63% |
| 8  | 8 | 197   | 99  | **37**  | −63% |
| 16 | 4 | 797   | 777 | **288** | −63% |
| 16 | 8 | 797   | 391 | **138** | −65% |
| 24 | 4 | 1 793 | 1769| **643** | −64% |
| 24 | 8 | 1 793 | 886 | **311** | −65% |

Before, a run of not-yet-loaded chunks at the head of the queue could consume the
whole per-tick budget and deliver nothing. After, only real work (send/generate)
spends the budget; not-ready chunks are cheaply skipped (bounded look-ahead), so
ready chunks behind them are delivered the same tick. Impact is largest exactly
when the server is under load (world-gen / disk).

### 3. `chunkLoadingQueue` leak — pruned on movement
`tick()` never pruned out-of-radius in-flight loads (only teleport / view-distance
change did). Residual stale entries after walking 200 chunks:

| VD | Before (no prune) | After (prune) |
|---:|---:|---:|
| 8  | 165 | **12** |
| 16 | 94  | **13** |
| 24 | 67  | **13** |

Each stale entry held a `CompletableFuture<IChunk>`; after the fix only the
genuinely in-flight loads remain.

## AI usage disclosure

| Model | What it did |
|-------|-------------|
|   GPT-5.6 Luna    |   Assisted with performances notes and performances tests          |

## Checklist

### DIFF
--- a/src/main/java/org/powernukkitx/level/PlayerChunkManager.java
+++ b/src/main/java/org/powernukkitx/level/PlayerChunkManager.java
@@ -1,5 +1,6 @@
 package org.powernukkitx.level;
 
+import org.jspecify.annotations.NonNull;
 import org.powernukkitx.Player;
 import org.powernukkitx.Server;
 import org.powernukkitx.entity.Entity;
@@ -26,12 +27,10 @@ import java.util.concurrent.TimeoutException;
 @Slf4j
 public final class PlayerChunkManager {
 
-
     /**
      * Chunks closer than this distance to the player are always considered to be in the field of view.
      */
     private static final double MIN_FOV_CHECK_DISTANCE = 4.0;
-
     /**
      * Timeout for asynchronously loading a chunk before retrying or generating it, in microseconds.
      */
@@ -98,6 +97,13 @@ public final class PlayerChunkManager {
     private final LongOpenHashSet pruneScratch;
     private long lastLoaderChunkPosHashed = Long.MAX_VALUE;
 
+    private static final int RESYNC_INTERVAL_TICKS = 10;
+    private static final int RESYNC_BATCH = 6;
+    private final LongArrayList resyncOrder = new LongArrayList();
+    private int resyncCursor = 0;
+    private int resyncCooldown = RESYNC_INTERVAL_TICKS;
+    private boolean resyncArmed = false;
+
     public PlayerChunkManager(Player player) {
         this.player = player;
         this.sentChunks = new LongOpenHashSet();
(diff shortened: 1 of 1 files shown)


sircurse:
We are implementing subchunk and refactoring all the leveldb along with the chunk streaming. Please wait until it is complete, since this class was completely refactored already.

PR #3107, opened 2026-08-28
fix: custom item on clear command

iYozemMc:
## What does this PR do?

Fix when you clear a custom item

## Why?

BugFix

## Type of change

- [X] Bug Fix
- [ ] New Feature
- [ ] Performance
- [ ] Documentation

## How was this tested?

- **Server build tested:** 4f30b91bb
- **Bedrock client version:** Latest
- **Plugins loaded:** Custom Plugin 

1. join the server and give you custom item
2. Type /clear @s <custom item> and /clear @s <custom item> 0 5 (for clear 5 time)

None

## Performance notes

N/A

## AI usage disclosure

| Model | What it did |
|-------|-------------|
|       |             |

- [x] This PR description and any review replies are written by me, not generated

## Checklist


xRookieFight (approved):

tunikakeks (approved):

xRookieFight (commented):
Good catch on the root cause - `setDamage(-1)` leaving meta at 65535 meant the filter could never match anything. Three notes:

**Line 49-52** - `checkData = true` assumes the `setDamage` above it landed. For tools and armour it goes through `toolSetDamage`/`wearableSetDamage`, which fire a cancellable `ItemWearEvent` and leave the meta alone if a plugin cancels. The item stays at meta 0 but we still compare damage, so `/clear @s diamond_sword 3` would clear undamaged swords instead. Building the matcher with `Item.get(item.getId(), data)` sidesteps it.

**Line 165** - the `inventory.setItem(0, ...)` → `offhand.setItem(0, ...)` change is correct (the old line dropped the offhand item into hotbar slot 0), but it's a different bug from the one in the title. Worth a sentence in the description so it doesn't look like a stray edit.

**Lines 111, 135, 172** - those `return 0`s exit the whole target loop, so `/clear @a stone` gives up as soon as one player has none. Pre-existing, but you're already editing every one of these branches, so it's cheap to fix here if you want it.

sircurse:
PR into beta pls.

PR #3146, opened 2026-09-05
fix(entity) : apply attack exhaustion to all mobs

NaySurGithub:
## What does this PR do?

It apply the attack exhaustion to every mob and with the vanilla value.

## Why?

It match vanilla behaviour.

The exhaustion was applied in `Player.attack(EntityDamageEvent)` in `Player.java`. That method is the one of the entity that **receive** the damage, and it only exist on `Player`, so the code only run when the victim is a player. Hitting a zombie, a cow or a villager was costing no hunger at all, only PvP was.

Last thing, `EntityDamageByChildEntityEvent extends EntityDamageByEntityEvent`, so an arrow shot by a player was passing the `instanceof EntityDamageByEntityEvent` check and was exhausting the shooter. In vanilla only the melee attack cost exhaustion, not the projectiles.

So the block is moved to the melee attack path in `InventoryTransactionHandler`, after a successful `target.attack()`, guarded with `target instanceof EntityLiving` and the survival/adventure check that the rest of that handler already use.

## Type of change

- [x] Bug Fix
- [ ] New Feature
- [ ] Performance
- [ ] Documentation

## How was this tested?

- **Server build tested:** 64e3c21e1
- **Bedrock client version:**
- **Plugins loaded:** none

None

## Performance notes

N/A

## AI usage disclosure

| Model | What it did |
|-------|-------------|
| Claude Opus 5 | Found the bug, wrote the diff, the commit message and a first draft of this description, reversed the minecraft bedrock exhaustion constants |

- [x] The disclosure above covers **all** AI use in this PR - code, tes config, and research

## Checklist

- [x] My contribution is licensed under LGPL-3.0, and any third-party c

### DIFF
--- a/src/main/java/org/powernukkitx/network/process/handler/InventoryTransactionHandler.java
+++ b/src/main/java/org/powernukkitx/network/process/handler/InventoryTransactionHandler.java
@@ -267,6 +267,9 @@ public class InventoryTransactionHandler implements PacketHandler<InventoryTrans
                     living.postAttack(player);
                 }
             }
+            if (target instanceof EntityLiving && (player.isSurvival() || player.isAdventure())) {
+                player.getFoodData().exhaust(0.3);
+            }
             if (item instanceof ItemMace mace) {
                 mace.onPostAttack(target, itemDamage);
             }
--- a/src/main/java/org/powernukkitx/network/process/handler/InventoryTransactionHandler.java
+++ b/src/main/java/org/powernukkitx/network/process/handler/InventoryTransactionHandler.java
@@ -268,7 +268,7 @@ public class InventoryTransactionHandler implements PacketHandler<InventoryTrans
                 }
             }
             if (target instanceof EntityLiving && (player.isSurvival() || player.isAdventure())) {
-                player.getFoodData().exhaust(0.3);
+                player.getFoodData().exhaust(0.1);
             }
             if (item instanceof ItemMace mace) {
                 mace.onPostAttack(target, itemDamage);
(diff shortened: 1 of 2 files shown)


src/main/java/org/powernukkitx/network/process/handler/InventoryTransactionHandler.java
```diff
@@ -267,6 +267,9 @@ private void handleUseItemOnEntity(@NotNull PlayerHandle playerHandle, @NotNull
                     living.postAttack(player);
                 }
             }
+            if (target instanceof EntityLiving && (player.isSurvival() || player.isAdventure())) {
+                player.getFoodData().exhaust(0.3);
```
Buddelbubi: According to the minecraft wiki, its 0.1. There is no difference between bedrock and java here

Buddelbubi (changes requested):

src/main/java/org/powernukkitx/network/process/handler/InventoryTransactionHandler.java
NaySurGithub: i literraly checked in minecraft bedrock code, i can test in bds... with a behavior pack or smthing like that

NaySurGithub: The wiki table does not split Java and Bedrock on that row, and the value it gives is the Java one. I checked in the Bedrock client (1.16.201): Player::attack calls Player::causeFoodExhaustion(0.3f) after a successful hit on a mob.
 The threshold is the same in both editions, so it is not a rescaled 0.1 - ExhaustionAttributeDelegate::tick drains when the value reaches mMaxExhaustionValue, and the Player constructor sets it to 4.0f, same as PNX uses.
 It is testable without the binary though, so I will just measure it: saturation at 0, full food bar, hit a mob and count. 0.1 means the bar drops after 40 hits, 0.3 means 14. I will post the result.

PleaseInsertNameHere (approved):

Buddelbubi (approved):

PR #3192, opened 2026-09-06
fix: snapshot actor data before sending to avoid ConcurrentModificationException

iYozemMc:
## What does this PR do?

Send a copy of the actorDataMap in the entity packets instead of the shared map

## Why?

The live map was being serialized on the Netty I/O thread while the server thread was modifying it -> ConcurrentModificationException (random crashes when a player)

## Type of change

- [X] Bug Fix
- [ ] New Feature
- [ ] Performance
- [ ] Documentation

## How was this tested?

- **Server build tested:** e6ad45529
- **Bedrock client version:** 26.45
- **Plugins loaded:** none

1.

Note: this is a race condition so it can't be reproduced 100% reliably

None

## Performance notes

N/A

## AI usage disclosure

| Model | What it did |
|-------|-------------|
|       |             |

## Checklist

- [X]  Commit messages follow Conventional Commits

### DIFF
--- a/src/main/java/org/powernukkitx/Player.java
+++ b/src/main/java/org/powernukkitx/Player.java
@@ -1515,7 +1515,7 @@ public class Player extends EntityHuman implements CommandSender, ChunkLoader, I
           mode update packet.
          */
         this.setGamemode(this.gamemode, false, null, true);
-        this.sendData(this.hasSpawned.values().toArray(Player.EMPTY_ARRAY), actorDataMap);
+        this.sendData(this.hasSpawned.values().toArray(Player.EMPTY_ARRAY), null);
         this.sendAttributes();
         this.spawnToAll();
         Arrays.stream(this.level.getEntities()).filter(entity -> entity.getViewers().containsKey(this.getLoaderId()) && entity instanceof EntityBoss).forEach(entity -> ((EntityBoss) entity).addBossbar(this));
--- a/src/main/java/org/powernukkitx/Player.java
+++ b/src/main/java/org/powernukkitx/Player.java
@@ -1515,7 +1515,7 @@ public class Player extends EntityHuman implements CommandSender, ChunkLoader, I
           mode update packet.
          */
         this.setGamemode(this.gamemode, false, null, true);
-        this.sendData(this.hasSpawned.values().toArray(Player.EMPTY_ARRAY), null);
+        this.sendData(this.hasSpawned.values().toArray(Player.EMPTY_ARRAY));
         this.sendAttributes();
         this.spawnToAll();
         Arrays.stream(this.level.getEntities()).filter(entity -> entity.getViewers().containsKey(this.getLoaderId()) && entity instanceof EntityBoss).forEach(entity -> ((EntityBoss) entity).addBossbar(this));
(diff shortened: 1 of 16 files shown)


tunikakeks (approved):

Buddelbubi:
Wouldn't it be better to make this.actorDataMap synchronized everywhere?
Or make it atomic?

sircurse (approved):

PleaseInsertNameHere:
> Wouldn't it be better to make this.actorDataMap synchronized everywhere? Or make it atomic?

Yes, we should synchronize instead of creating a copy.

PleaseInsertNameHere (changes requested):
Please synchronize instead of creating copies.

iYozemMc:
I synchronized all access to actorDataMap as suggested, but kept a single snapshot at send time: the map is serialized later on the Netty thread, so synchronization alone can't prevent the CME without a copy at that boundary

src/main/java/org/powernukkitx/entity/EntityLiving.java
```diff
@@ -962,7 +962,7 @@ private void setBlockingFlags(boolean blocking, boolean transitionBlocking) {
 
         this.setDataFlags(ext);
 
-        sendData(this.hasSpawned.values().toArray(Player.EMPTY_ARRAY), this.actorDataMap);
+        sendData(this.hasSpawned.values().toArray(Player.EMPTY_ARRAY), null);
```
Buddelbubi: Wth null?

src/main/java/org/powernukkitx/Player.java
```diff
@@ -1515,7 +1515,7 @@ public void onPlayerLocallyInitialized() {
           mode update packet.
          */
         this.setGamemode(this.gamemode, false, null, true);
-        this.sendData(this.hasSpawned.values().toArray(Player.EMPTY_ARRAY), actorDataMap);
+        this.sendData(this.hasSpawned.values().toArray(Player.EMPTY_ARRAY), null);
```
Buddelbubi: Wth null?

Buddelbubi (changes requested):

src/main/java/org/powernukkitx/entity/EntityLiving.java
iYozemMc: null means "send the whole map", it goes through snapshotActorData() instead of being used as a delta. These sites used to pass this.actorDataMap directly, which is what caused the CME. Agreed it reads badly though: switched to the existing sendData(Player[]) overload that does the same thing, plus javadoc on the null semantics

src/main/java/org/powernukkitx/Player.java
iYozemMc: null means "send the whole map", it goes through snapshotActorData() instead of being used as a delta. These sites used to pass this.actorDataMap directly, which is what caused the CME. Agreed it reads badly though: switched to the existing sendData(Player[]) overload that does the same thing, plus javadoc on the null semantics

xRookieFight:
@codex

xRookieFight:
@codex review

xRookieFight:
@codex review

xRookieFight:
@codex review

PR #3235, opened 2026-09-19
feat: levelDB Parity, new features, many bug fixes.

sircurse:
.
.
### DO NOT TEST THIS PR IN PRODUCTION, THERE ARE IRREVERSIBLE CHANGES TO SERVER DATA, WORLD DATA!!!
.
.

## What does this PR do?

This PR brings PowerNukkitX world persistence and several related runtime systems substantially in line with modern Bedrock Dedicated Server behavior.

Reaching real LevelDB parity required more than changing serializers. Many persisted values are part of larger BDS runtime systems, so this PR also improves entity identity, chunks, SubChunk streaming, client session initialization, lighting, structures, world generation, block entities, ticking areas, scoreboards, position tracking, vibrations and other gameplay systems.

Legacy PNX data is migrated to the new canonical formats, while runtime code operates only on the current representation.

## Major features

* **BDS-compatible actor persistence**

  * `actorprefix` / `digp` storage.
  * Persistent `ActorUniqueID`.
  * Actor digest/version handling.
  * Actor links, passengers and cross-chunk movement.
  * Limbo entity handling.
  * Duplicate ActorUniqueID protection.

* **Server-global LevelDB storage**

  * Player-related data.
  * Scoreboards.
  * Position tracking/lodestones.
  * Dynamic Properties.
  * ActorUniqueID allocation and other server-scoped state.

* **Full migration framework**

  * Migrates legacy PNX and BDS data where required.
  * Covers players, actors, block entities, chunks, structures, scoreboards, position tracking, Dynamic Properties and world-global storage.
  * Existing canonical data wins on conflicts.

* **Expanded BDS chunk parity**

  * Data3D/biomes.
  * BiomeState.
  * heightmaps and rain heightmaps.
  * random and scheduled ticks.
  * Border Blocks.
  * AABBVolumes.
  * LevelChunkMetaData and metadata dictionaries.
  * chunk finalization state.
  * blending-related records.
  * actor digest references.

* **BDS-style SubChunk streaming**

  * Implements `SubChunkRequest` / `SubChunk` terrain streaming.
  * Client-requested individual SubChunks instead of relying only on complete chunk transmission.
  * Correct success, all-air, out-of-bounds, missing-chunk and wrong-dimension results.
  * Per-SubChunk block entities.
  * BDS heightmap and render-heightmap data in SubChunk responses.
  * Integrates Anti-Xray serialization with the SubChunk path.

* **Client blob-cache support**

  * Negotiates client cache capability.
  * Uses content hashes for reusable SubChunk terrain blobs.
  * Handles client cache hits, misses and acknowledgements.
  * Tracks cache state per connection.
  * Limits concurrent transfers according to network load.
  * Correctly bypasses cached data when chunk contents require a fresh transmission.

* **BDS-aligned client connection and world-entry state machine**

  * Explicit connection progression through network settings, login, authentication, encryption, resource packs, pre-spawn and chunk streaming.
  * Packets are processed only in the appropriate session state.
  * Authentication work is moved away from the network event loop while preserving login validation order.
  * Initial `StartGame`, actor identity and runtime identity follow the expected Bedrock separation.
  * Initial spawn now follows the Bedrock respawn/search/ready sequence instead of immediately forcing the player into the world.
  * Player spawning waits for the required initial chunk area to become available.
  * Client local initialization is handled separately from server login/spawn state.
  * Initial chunk radius and later client-requested view distance are handled as separate stages.
  * Improves connection stability by preserving the expected packet/state ordering instead of treating login as a single boolean state.

* **New chunk generation lifecycle**

  * Separates runtime generation state from persisted finalization state.
  * Adds generation dependencies and neighbor-aware generation.
  * Improves chunk build/send ordering.

* **New lighting system**

  * Initial SubChunk lighting and runtime relighting.
  * Correct block/sky-light persistence.
  * `SubChunkLightData`.
  * Neighbor-aware light propagation.
  * Replaces the old `LightPopulated`-based lifecycle.

* **BDS structure placement parity**

  * Correct random-spread placement, salts, spacing and RNG.
  * Strongholds, monuments, villages, Ancient Cities, trial chambers, trail ruins, Nether structures and others now use BDS-oriented placement logic.

* **Structure AABB volumes and spawn overrides**

  * Persistent BDS structure bounds.
  * Structure-specific mob spawning, including monuments, fortresses, swamp huts and outposts.

* **Ancient City generation fixes**

  * Correct BDS placement salt and triangular spread.
  * Correct Deep Dark biome sampling.
  * Correct `city_anchor` jigsaw start.
  * Correct generation bounds, terrain adaptation and neighboring-piece handling.

* **End Dragon fight persistence**

  * Persistent Dragon Fight state.
  * Exit portal handling.
  * End gateways.
  * obsidian pillars/crystals.
  * dragon respawning.

* **Creaking / Creaking Heart parity**

  * Persistent Heart ↔ Creaking association.
  * Correct home/dimension tracking.
  * observed/twitching/crumbling states.
  * damage behavior and resin interaction.
  * movement/AI behavior.
  * particles, sounds and lifecycle handling.

* **Persistent vibration / GameEvent system**

  * BDS GameEvent IDs and frequencies.
  * persisted vibration listeners.
  * projectile/source ActorUniqueIDs.
  * improved Sculk Sensor, Calibrated Sculk Sensor, Shrieker and Warden behavior.

* **Biome/snow persistence**

  * BDS BiomeState.
  * precipitation/rain-height handling.
  * persistent snow RNG/state.
  * snowlogging behavior.

* **LevelDB ticking areas**

  * Replaces `tickingarea.json`.
  * UUID identity, dimensions, preload state and per-world limits.

* **LevelDB scoreboards**

  * Persistent ScoreboardIds and identities.
  * ActorUniqueID-based player/entity scorers.

* **LevelDB position tracking**

  * Replaces the PNX `.pnt` storage.
  * Canonical lodestone/compass tracking records and migration.

## Important bug fixes and parity improvements

* **Double chests now correctly remain paired across chunk boundaries.**

  * Pair leader/state is persisted correctly.
  * Loading or unloading either side no longer incorrectly destroys the pairing.
  * Inventory, lid state and open/close behavior remain synchronized.

* Fixed Ancient Cities generating with incorrect placement, biome and jigsaw behavior.

* Fixed Creaking and Creaking Heart behavior and persistence.

* Fixed client world-entry ordering so the player is not considered fully spawned before the required initialization, respawn and chunk stages complete.

* Fixed differences between persistent ActorUniqueID and runtime entity ID in `StartGame`, actor links and other network paths.

* Fixed SubChunk terrain responses, including block entities, all-air sections, dimensions and heightmaps.

* Fixed structure placement at negative coordinates and multiple incorrect structure RNG sequences.

* Fixed Stronghold, Ocean Monument, Nether Fortress/Bastion and Mineshaft placement behavior.

* Fixed structure-specific mob spawning through persisted AABB volumes.

* Fixed entity persistence using runtime IDs/UUID assumptions instead of persistent ActorUniqueIDs.

* Fixed riding/passenger relationships after restart.

* Fixed player inventory, armor, offhand, main hand and Ender Chest persistence to use canonical Bedrock layouts.

* Fixed unknown Bedrock item/NBT fields being lost when PNX rewrites data.

* Fixed Vault player identities and other actor references.

* Fixed scoreboard identities and persistent ScoreboardIds.

* Fixed lodestone position-tracking persistence.

* Fixed BDS heightmap semantics and the previous off-by-one interpretation.

* Fixed rain/precipitation height being incorrectly tied to the normal heightmap.

* Fixed chunk sections being considered empty when the second block layer still contained blocks.

* Fixed several asynchronous world-generation features accessing invalid neighboring sections.

* Fixed lighting initialization and propagation around chunk/SubChunk boundaries.

* Fixed vibration persistence and Sculk/Warden event handling.

* Fixed End Gateway / End Dragon fight persistence and related End behavior.

* Fixed several block entities whose stored NBT differed from BDS.

* Fixed item storage format, including canonical empty items and block-item state preservation.

### Entity identity

This is the most important plugin-facing change.

PNX entity UUIDs are **no longer generated automatically**.

Entity identity now follows Bedrock:

* `entity.uniqueId()` / `entity.uniqueIdLong()` = persistent BDS `ActorUniqueID`.
* `Entity#getUniqueId()` is deprecated and may return `null` for newly created non-player entities.
* UUID is now optional PNX custom data and is **not** the canonical actor identity.

Plugins previously using `getUniqueId()` to identify entities should migrate to:

* `Entity#uniqueId()`
* `Entity#uniqueIdLong()`

Plugins that specifically require a UUID for their own data must now explicitly create or assign one:

* `Entity#generateUUID()`
* `Entity#setUUID(UUID)`
* `Entity#getUUID()`
* `Entity#removeUUID()`

In other words, plugins must no longer assume that every non-player entity automatically has a PNX-generated UUID.

### Removed classes

* `ChunkConversion`
* `BDSEntityTranslator`
* `JSONTickingAreaStorage`
* `BedrockMigrationService`

These are replaced by canonical runtime storage and the new migration framework.

### Chunk API

`IChunk` has new required APIs for:

* generation/finalization state;
* biome sections;
* height/rain/render heightmaps;
* random updates;
* metadata;
* AABB volumes;
* BiomeState;
* Border Blocks.

The old Border Block column-cache API was removed.

`IChunk.VERSION` changes from `40` to `42`.

`ChunkState` is deprecated in favor of:

* `ChunkGenerationState`
* `ChunkFinalizationState`

`IChunkBuilder` also gains the corresponding new storage APIs.

### ChunkSection

`ChunkSection`'s record layout changes from separate block/sky-light arrays to `SubChunkLightData`.

### Ticking areas

`TickingAreaStorage` changes from name-based:

`Map<String, TickingArea>`

to UUID-based:

`Map<UUID, TickingArea>`

and gains Level/world-specific persistence APIs.

`TickingAreaManager` gains new abstract world/UUID/preload operations.

### Position tracking

`PositionTrackingService(File)` becomes `PositionTrackingService(DB)`.

`PositionTrackingStorage` no longer uses file-backed `.pnt` storage and its previous file/start-index constructors and related methods are removed.

### Scoreboards

`EntityScorer(UUID)` becomes ActorUniqueID-based.

`PlayerScorer` is internally ActorUniqueID-based; UUID compatibility APIs are deprecated.

### Block entities / entities

Notable signature changes include:

* `BlockEntityBeehive.Occupant(...)`
* `BlockEntityStructureBlock#updateSetting(Player, packet)`
* Vault rewarded/connected player APIs: `String` → `long ActorUniqueID`
* `WardenSniffExecutor` constructor.
* Custom entity `Attack` record/API replaced by `AttackComponent` / `AttackDamageComponent`.
* `EntityHumanType` no longer directly implements `InventoryHolder`.
* `EntityPillager` and `EntityVindicator` no longer implement `EntityWalkable`.

### Other API changes

* `RandomSourceProvider` gains `nextExclusiveInt(int)`.
* `VibrationManager` gains `tick()`.
* `VibrationEvent` record now also contains source/projectile-owner ActorUniqueIDs.
* `ItemHelper.write(Item, Integer)` is replaced by primitive/int and preservation-aware overloads.
* `LevelDBProvider.writeLevelDat(...)` signature changes.
* several structure-generation extension points changed as part of the new placement/generation lifecycle.

## Deprecated APIs

Deprecated for 3.1.0:

* `Entity#getUniqueId()`
* legacy `Entity#entityUniqueId`
* `EntityFishingHook#setTarget(long)`
* `EntityItem#setOwner(String)`
* `EntityItem#getThrower()`
* `EntityItem#setThrower(String)`
* `PlayerScorer(UUID)`
* `PlayerScorer#getUuid()`
* `ChunkState`
* legacy `IChunk` generation-state helpers:

  * `getChunkState()`
  * `setChunkState(...)`
  * `isGenerated()`
  * `isPopulated()`
  * `isFinished()`
  * `setGenerated()`
  * `setPopulated()`
* legacy direct lighting APIs:

  * `populateSkyLight()`
  * `setLightPopulated(...)`
* `IChunkBuilder#state(ChunkState)`
* corresponding `Chunk` / `UnsafeChunk` legacy state and lighting methods.
* `VibrationType.INSTRUMENT_PLAY`

## Compatibility warning

This PR performs migrations of world and server data into new canonical storage formats.

World/server data opened and migrated by this version should **not be opened again with older PNX versions**.

Plugins relying on low-level chunk APIs, custom ticking-area/position-tracking implementations, or especially the old automatically generated entity UUID must be updated.

## Type of change

- [X] Bug Fix
- [X] New Feature
- [X] Performance
- [X] Refactor (no behaviour change)
- [ ] Documentation
- [X] Build / CI / dependencies

## How was this tested?

- **Server build tested:** 332fddf9f78957c875b2b26629950eef7b6f173f
- **Bedrock client version:** 26.51
- **Plugins loaded:** none

1.
2.

- [X] I added tests for the new logic, or explained below why that isn't practical

## AI usage disclosure

| Model | What it did |
|-------|-------------|
|   codex    |     heavily BDS vs PNX inspection on leveldb, format, data, schema, correlation        |

## Checklist


xRookieFight:
@codex review

xRookieFight:
@codex review

xRookieFight:
@codex

src/main/java/org/powernukkitx/network/RakNetNetworkMetrics.java
```diff
@@ -0,0 +1,94 @@
...
+     * @return the requested value
+     */
+    public NetworkLoad getNetworkLoad() {
+        this.updateBandwidthEstimate();
+        final int queuedBytes = this.queuedPacketBytes;
+
+        if (queuedBytes <= 0) return NetworkLoad.LOW;
+
+        final int maxBps = this.approximateMaxBps;
+        if (maxBps <= 0) return NetworkLoad.HIGH;
```
PleaseInsertNameHere, line 58: zero bandwith estimate is classified as HIGH load

src/main/java/org/powernukkitx/level/PlayerChunkManager.java
```diff
@@ -222,7 +479,121 @@ private void removeOutOfRadiusChunks() {
...
+
+    /**
+     * Requeues a stale SubChunk transaction when the chunk is still eligible for recovery.
+     *
+     * @param chunkX chunk X
+     * @param chunkZ chunk Z
+     * @return whether recovery was scheduled
+     */
+    @ApiStatus.Internal
+    public synchronized boolean tryRecoverStaleSubChunkTransaction(int chunkX, int chunkZ) {
```
PleaseInsertNameHere, line 531: No callers; recoveryChunks is never populated and `shouldBypassChunkCache()` always returns false

src/main/java/org/powernukkitx/level/PlayerChunkManager.java
```diff
@@ -222,7 +479,121 @@ private void removeOutOfRadiusChunks() {
...
+        if (baseBudget == 0) return 0;
+
+        final int activePlayerCount = player.getLevel().getPlayers().size();
+        if (activePlayerCount <= 0) {
+            return baseBudget;
         }
+
+        final int dividedBudget = baseBudget >= activePlayerCount ? baseBudget / activePlayerCount : 0;
+        return dividedBudget >= 2 ? dividedBudget : 1;
     }
```
PleaseInsertNameHere, line 597: This replaces the configured per-tick chunk send count with a hardcoded network-load table - real throughput is capped far below the configured value

src/main/java/org/powernukkitx/network/process/cache/ClientBlobCacheManager.java
```diff
@@ -0,0 +1,477 @@
...
+     * @param session value for this API
+     * @return the requested value
+     */
+    public static TransferBuilder tryStartTransfer(BedrockServerSession session) {
+        final SessionState state = SESSIONS.get(session);
+        if (state == null) return null;
+
+        synchronized (state) {
+            final int maximumTransfers = getMaximumConcurrentTransfers(session);
+            if (state.activeTransfers.size() > maximumTransfers) return null;
```
PleaseInsertNameHere, line 172: should be `>=`

src/main/java/org/powernukkitx/network/process/cache/ClientBlobCacheManager.java
```diff
@@ -0,0 +1,477 @@
...
+        }
+
+        @Override
+        public void close() {
+            commit();
+        }
+    }
+
+    private static final class SessionState {
+        private final ConcurrentHashMap<Long, byte[]> pending = new ConcurrentHashMap<>();
```
PleaseInsertNameHere, line 375: Only pruned on client acknowledgment; if a client never acknowledges, server memory grows without limit

src/main/java/org/powernukkitx/network/process/handler/SubChunkRequestHandler.java
```diff
@@ -0,0 +1,283 @@
...
+
+            player.sendPacketImmediately(responsePacket);
+            return;
+        }
+
+        final DimensionData dimensionData = level.getDimensionData();
+        final boolean cacheSupported = !shouldBypassCacheForRequest(player, centerPos, packet) && ClientBlobCacheManager.isEnabled(player.getSession());
+
+        if (cacheSupported) {
+            cacheTransfer = ClientBlobCacheManager.startTransfer(player.getSession());
```
PleaseInsertNameHere, line 70: Concurrency check missing - limit on concurrent blob transfers is never applied

src/main/java/org/powernukkitx/Player.java
```diff
@@ -3946,18 +4126,43 @@ public boolean kick(PlayerKickEvent.Reason reason, String reasonString, String d
...
+    /**
+     * Stores the chunk radius requested by the client and applies its server-clamped value.
+     *
+     * @param distance client-requested chunk radius
+     */
+    public void setClientRequestedChunkRadius(int distance) {
+        this.clientRequestedChunkRadius = distance;
+        int clampedDistance = Math.max(MIN_CLIENT_REQUESTED_CHUNK_RADIUS, Math.min(distance, this.server.getViewDistance()));
+        this.applyViewDistance(clampedDistance);
+    }
```
PleaseInsertNameHere, line 4141: If `gameplaySettings.viewDistance` is below 5, a client request raises the view distance above the configured limit

src/main/java/org/powernukkitx/level/format/leveldb/LevelDBChunkSerializer.java
```diff
@@ -70,93 +72,220 @@ public void serialize(WriteBatch writeBatch, IChunk chunk) {
...
+    private void serializeBorderBlocks(WriteBatch writeBatch, UnsafeChunk chunk) {
+        byte[] key = LevelDBKeyUtil.BORDER_BLOCKS.getKey(chunk.getX(), chunk.getZ(), chunk.getDimensionData());
+        byte[] data = new byte[256];
+        int count = 0;
+
+        outer:
+        for (int localX = 0; localX < 16; localX++) {
+            for (int localZ = 0; localZ < 16; localZ++) {
+                if (!chunk.hasBorderBlock(localX, localZ)) continue;
+                if (count >= 255) break outer;
```
PleaseInsertNameHere, line 222: A chunk in which all 256 columns are border blocks loses one column on every save

src/main/java/org/powernukkitx/level/format/ChunkSection.java
```diff
@@ -64,15 +70,133 @@ public BlockState getBlockState(int x, int y, int z, int layer) {
...
+            final boolean layer0Empty = blockLayer[0].isEmpty();
+            final boolean layer1Empty = blockLayer[1].isEmpty();
+
+            if (layer0Empty && layer1Empty) {
+                return lighting.installCombinedLightPropertiesCache(created);
+            }
+
+            if (layer1Empty) {
+                for (int blockIndex = 0; blockIndex < SIZE; blockIndex++) {
+                    created[blockIndex] = compactCombinedLightProperties(blockLayer[0].getLightingProperties(blockIndex));
```
PleaseInsertNameHere, line 120: The fast paths in `getCombinedLightingProperties()` feed the raw palette value to `compactCombinedLightProperties()`, skipping the `Math.min(15, filter)` clamp that `calculateCombinedLightProperties()` applies

src/main/java/org/powernukkitx/level/format/Chunk.java
```diff
@@ -80,86 +103,95 @@ public class Chunk implements IChunk {
...
+        this.biomeSections = createBiomeSections(levelProvider.getDimensionData().getChunkSectionCount());
         this.heightMap = new short[256];
+        this.renderHeightMap = new short[256];
+        this.rainHeightMap = ChunkRainHeightMap.create();
         this.entities = new Long2ObjectNonBlockingMap<>();
         this.tiles = new Long2ObjectNonBlockingMap<>();
         this.tileList = new Long2ObjectNonBlockingMap<>();
-        this.blockUpdateScheduler = new BlockUpdateScheduler(this, levelProvider.getLevel().getCurrentTick());
+        this.blockUpdateScheduler = new BlockUpdateScheduler(this, levelProvider.getCurrentTick());
+        this.randomBlockUpdateScheduler = new RandomBlockUpdateScheduler(this, 0);
```
PleaseInsertNameHere, line 136: Seeded with lastTick `0` instead of the provider's current tick. First tick() after an entry is queued can iterate tens of thousands of empty ticks.

src/main/java/org/powernukkitx/level/format/Chunk.java
```diff
@@ -80,86 +103,95 @@ public class Chunk implements IChunk {
...
+        this.biomeSections = biomeSections;
         this.heightMap = heightMap;
+        this.renderHeightMap = new short[256];
+        this.rainHeightMap = ChunkRainHeightMap.create();
         this.entities = new Long2ObjectNonBlockingMap<>();
         this.tiles = new Long2ObjectNonBlockingMap<>();
         this.tileList = new Long2ObjectNonBlockingMap<>();
-        this.blockUpdateScheduler = new BlockUpdateScheduler(this, levelProvider.getLevel().getCurrentTick());
+        this.blockUpdateScheduler = new BlockUpdateScheduler(this, levelProvider.getCurrentTick());
+        this.randomBlockUpdateScheduler = new RandomBlockUpdateScheduler(this, 0);
```
PleaseInsertNameHere, line 181: Seeded with lastTick `0` instead of the provider's current tick. First tick() after an entry is queued can iterate tens of thousands of empty ticks.

PleaseInsertNameHere (changes requested):

Zephyr-OP:
05:38:36 [main] [WARN] destination stack size bigger than the max stack size!
05:38:39 [main] [WARN] transfer an item to a slot that has a different item is not allowed
05:38:40 [main] [WARN] transfer an item to a slot that has a different item is not allowed
05:38:40 [main] [WARN] transfer an item to a slot that has a different item is not allowed
05:38:40 [main] [WARN] transfer an item to a slot that has a different item is not allowed
05:38:43 [main] [WARN] transfer an item to a slot that has a different item is not allowed
05:38:43 [main] [WARN] transfer an item that has not enough count is not allowed. Expected: 63, Actual: 64
05:38:44 [main] [WARN] transfer an item to a slot that has a different item is not allowed
05:38:45 [main] [WARN] transfer an item to a slot that has a different item is not allowed
05:38:45 [main] [WARN] transfer an item to a slot that has a different item is not allowed
05:38:58 [main] [WARN] transfer an item to a slot that has a different item is not allowed
05:39:17 [main] [WARN] cannot throw more items than the current amount!

ever since trying this my inventory is sometimes out of sync for like cushions and stuff I do not fully know if it's cause of this or not but I grabbed 64 cushions and you should only be able to grab 16.

Zephyr-OP:
@sircurse  Another thing to fix is that custom entities no longer get their textures for some reason.

src/main/java/org/powernukkitx/level/PlayerChunkManager.java
sircurse: This is intentional. The chunk streaming budget is meant to adapt to the current network load rather than directly follow the configured static value. The configured value was the previous PNX behavior, while this logic aligns chunk delivery to the client with BDS behavior.

src/main/java/org/powernukkitx/network/process/cache/ClientBlobCacheManager.java
sircurse: The strict > is intentional here. Reaching the configured transfer limit is still allowed; throttling starts only once the active transfer count exceeds it.
Changing this to >= would introduce an off-by-one difference in the intended transfer behavior.

src/main/java/org/powernukkitx/network/process/handler/SubChunkRequestHandler.java
sircurse: This is intentional. The transfer pressure limit applies to the server-driven chunk publisher, while SubChunk responses are client-requested transactions and follow a separate path.
Applying the same concurrency gate here would incorrectly throttle client SubChunk requests.

src/main/java/org/powernukkitx/Player.java
sircurse: This is intentional. The client request is still capped by the server-configured view distance, so it cannot increase the radius beyond the server setting.
The only exception is the minimum radius of 5: if the server is configured below 5, the effective client-visible radius is still raised to 5 to match BDS behavior.

src/main/java/org/powernukkitx/level/format/Chunk.java
sircurse: This is intentional. The random tick scheduler keeps its own persisted timeline rather than using the level runtime tick. When RandomTicks data is loaded, its stored tick value is restored into the scheduler, so seeding it from the provider tick here would mix two different timelines and could change persisted scheduling behavior.
Also, when there is no pending work, the scheduler advances its internal tick without iterating through all previous ticks.

src/main/java/org/powernukkitx/level/format/Chunk.java
sircurse: Same reasoning applies in this constructor. The 0 initialization is deliberate because the random tick scheduler is restored from its own persisted currentTick value, not from the level provider's runtime tick.
Using levelProvider.getCurrentTick() here would alter the RandomTicks persistence semantics. The scheduler also avoids replaying an unbounded range of empty ticks when there is no queued work.

src/main/java/org/powernukkitx/network/RakNetNetworkMetrics.java
sircurse: This is intentional. The maxBps <= 0 check is only reached when there are already queued bytes; an idle connection returns LOW before this point.

When traffic is queued but a usable bandwidth estimate is not available yet, the connection is treated as high pressure rather than allowing a larger chunk-streaming budget. This matches the expected BDS network-pressure behavior.

src/main/java/org/powernukkitx/level/format/leveldb/LevelDBChunkSerializer.java
sircurse: This is an intentional format limitation rather than a serialization bug. The border-block format stores the entry count in a single byte, so it can represent at most 255 columns.
Supporting 256 entries would require changing the on-disk format rather than fixing the current serializer, and would break compatibility with the Bedrock format we are preserving.

PleaseInsertNameHere: Reasonable

src/main/java/org/powernukkitx/network/process/cache/ClientBlobCacheManager.java
PleaseInsertNameHere: With the strict `>` the possible maximum is always off-by-one.

When `size == maximumTransfers` passes the check, commit() on line 361 does activeTransfers.add(...) resulting in max + 1 active

src/main/java/org/powernukkitx/Player.java
PleaseInsertNameHere: I don't get why this should be BDS behavior. Let me give you a table of results:


| gameplaySettings.viewDistance | client asks | effective |
|--|--|--|
| 16 | 32 | 16 |
| 16 | 3 | 5 |
| 5 | 32 | 5 |
| 4 | anything | 5 |
| 3 | anything | 5 |
| 2 | anything | 5 |

A client asks for less, still gets five. Even if this is BDS behavior, I think in this case we shouldn't match it.

PleaseInsertNameHere: Furthermore, `setViewDistance()` from a plugin clamps to 3, while the client path yields 5

src/main/java/org/powernukkitx/level/PlayerChunkManager.java
sircurse: Latest commits added a dynamic budget, replacing the BDS's hardcoded and highly restrictive budget.

src/main/java/org/powernukkitx/network/process/cache/ClientBlobCacheManager.java
sircurse: The `limit + 1` observation is correct, but it is intentional rather than an off-by-one. Native BDS publisher stops only when `activeTransferCount > configuredLimit`, equality is explicitly allowed. The Java method name makes this look like an inclusive maximum, so I will clarify (added a comment) that it represents the native pressure threshold.

src/main/java/org/powernukkitx/network/process/cache/ClientBlobCacheManager.java
sircurse: This retention is intentional and was separately verified against BDS. Native active transfers keep strong ownership of advertised blob payloads until the client reports Found/Missing or the connection/manager is destroyed. PNX likewise clears session state on disconnect. A client that never resolves transfers can therefore retain data for the connection lifetime, adding a separate misbehaving-client memory bound could be useful hardening, but a TTL that drops unresolved blobs would be a deliberate divergence from BDS and could break a later valid Missing response.

src/main/java/org/powernukkitx/Player.java
sircurse: Ok, changed it, now the view-distance setting should remain the absolute upper bound, so I changed the clamp ordering accordingly.

src/main/java/org/powernukkitx/level/format/ChunkSection.java
sircurse: BlockLightProperties.packOf() now normalizes both light emission and light filter to the valid 0..15 range before packing them, that keeps compactCombinedLightProperties() as a cheap nibble compaction step.

xRookieFight:
@claude review

